PR #27, opened 2022-12-18
Add support to verify a signed message

0xmaayan:
Add function `signMessageAndVerify` in `WalletCore` to verify the signature  and prove a user controls a given private key.

### DIFF
--- a/packages/wallet-adapter-core/src/WalletCore.ts
+++ b/packages/wallet-adapter-core/src/WalletCore.ts
@@ -321,17 +321,20 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
           .message;
       // Verify that the bytes were signed using the private key that matches the known public key
       let verified = false;
-      // Remove the 0x prefix
+      // support for when address doesnt have hex prefix (0x)
       const currentAccountPublicKey = new HexString(this._account.publicKey);
-      // multi sig wallets
+
       if (Array.isArray(response.signature)) {
+        // multi sig wallets
         // TODO - implement multi sig wallets
-        console.log(response);
       } else {
         // single sig wallets
+
+        // support for when address doesnt have hex prefix (0x)
+        const signature = new HexString(response.signature);
         verified = sign.detached.verify(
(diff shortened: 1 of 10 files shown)


packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -307,4 +310,36 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
       throw new WalletNetworkChangeError(errMsg).message;
     }
   }
+
+  async signMessageAndVerify(message: SignMessagePayload): Promise<boolean> {
+    try {
+      this.isWalletExists();
```
hariria: Might be good to rename to doesWalletExist or walletExists

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -307,4 +310,36 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
+      if (!this._account) throw new Error("No account found!");
+      const response = await this._wallet?.signMessage(message);
+      if (!response)
+        throw new WalletSignMessageAndVerifyError("Failed to sign a message")
+          .message;
+      // Verify that the bytes were signed using the private key that matches the known public key
+      let verified = false;
+      // Remove the 0x prefix
+      const currentAccountPublicKey = this._account.publicKey.slice(2, 66);
```
hariria: What if they don't have hex prefix?

hariria: I think there's a verify method on AccountAddress in Aptos package

hariria (approved):
LGTM some nits

packages/wallet-adapter-core/src/WalletCore.ts
0xmaayan: thanks! I think there is a WIP validate method to add to the ts sdk, but was able to find a way to ensure this is a hex string and then remove its prefix to use it in the signature.verify method

PR #28, opened 2022-12-19
Support multisig wallet

scottphc:
Support multisig wallet, like https://github.com/hippospace/aptos-wallet-adapter/pull/69

Before we integrate with Blocto, we need to change the interface first.

### DIFF
--- a/packages/wallet-adapter-core/src/types.ts
+++ b/packages/wallet-adapter-core/src/types.ts
@@ -11,7 +11,8 @@ export type NetworkInfo = {
 
 export type AccountInfo = {
   address: string;
-  publicKey: string;
+  publicKey: string | string[];
+  minKeysRequired?: number
 };
 
 export interface AptosWalletErrorResult {
--- a/.changeset/dull-onions-rule.md
+++ b/.changeset/dull-onions-rule.md
@@ -1,0 +1,5 @@
+---
+"@aptos-labs/wallet-adapter-core": patch
+---
+
+Add multisig support for AccountInfo


0xmaayan:
@scottphc LGTM, thanks!

Could you also add it to the changelog so we can bump and publish a new version.
can follow https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/CONTRIBUTING.md#changesets

scottphc:
@0xmaayan added!

scottphc:
> @scottphc LGTM, thanks!
> 
> Could you also add it to the changelog so we can bump and publish a new version. can follow https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/CONTRIBUTING.md#changesets

@0xmaayan Added!

0xmaayan (approved):

packages/wallet-adapter-core/src/types.ts
```diff
@@ -11,7 +11,8 @@ export type NetworkInfo = {
 
 export type AccountInfo = {
   address: string;
-  publicKey: string;
+  publicKey: string | string[];
+  minKeysRequired?: number
```
kent-white, line 15: I've had this question for a while but why do we need `minKeysRequired`?

Shouldn't that be coming from on-chain?

If you set minKeysRequired coming back from the wallet an attacker could easily just pass "1"

scottphc: Where can I get minKeysRequired on chain? As I know, there is an Auth Key only. According to https://aptos.dev/concepts/accounts#multisigner-authentication, developers can verify it by using `sha3-256(p_1 | . . . | p_n | K | 0x01)` and then comparing with Auth Key on chain.

PR #29, opened 2022-12-19
Setup aptos-wallet-adapter-tailwind-ui

WEB3Zer0:


0xmaayan:
thanks @WEB3Zer0 , it looks good!

I think it might be a good idea to also include the `wallet.icon` on the wallet select modal and an indication whether the wallet is installed, and if not we can add a CTA and redirect them to the `wallet.url` - something like what [aptosnames.com](https://www.aptosnames.com/) has.

Also, it might help to add the new UI to the nextjs-example app so we can see how it looks

apps/nextjs-example/package.json
```diff
@@ -13,7 +13,9 @@
   "dependencies": {
     "@aptos-labs/wallet-adapter-core": "workspace:*",
     "@aptos-labs/wallet-adapter-react": "workspace:*",
+    "@aptos-labs/wallet-adapter-tailwind-ui": "workspace:*",
     "@martianwallet/aptos-wallet-adapter": "^0.0.2",
+    "@headlessui/react": "^1.7.7",
```
0xmaayan: dont think this is needed here, it should be on the `ui` package.

apps/nextjs-example/pages/index.tsx
```diff
@@ -1,10 +1,12 @@
 import { AptosClient, Types } from "aptos";
 import { useWallet } from "@aptos-labs/wallet-adapter-react";
-import { useState } from "react";
+import { Fragment, useState } from "react";
 import { ErrorAlert, SuccessAlert } from "../components/Alert";
 import dynamic from "next/dynamic";
 import Image from "next/image";
 import { useAutoConnect } from "../components/AutoConnectProvider";
+import { WalletDropdown } from "@aptos-labs/wallet-adapter-tailwind-ui";
+import { Menu, Transition } from '@headlessui/react';
```
0xmaayan: can remove it as it is being used on the `ui` package

apps/nextjs-example/pages/index.tsx
```diff
@@ -1,10 +1,12 @@
 import { AptosClient, Types } from "aptos";
 import { useWallet } from "@aptos-labs/wallet-adapter-react";
-import { useState } from "react";
+import { Fragment, useState } from "react";
```
0xmaayan: can remove it as it is being used on the `ui` package

apps/nextjs-example/pages/index.tsx
```diff
@@ -28,6 +30,8 @@ export default function App() {
     signTransaction,
     signMessage,
     signMessageAndVerify,
+    wallets,
+    connect
```
0xmaayan: can remove both as it is not being used on this file

packages/wallet-adapter-tailwind-ui/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "@aptos-labs/wallet-adapter-tsconfig": "workspace:*",
+    "@types/react": "^18.0.17",
+    "@types/react-dom": "^18.0.6",
+    "eslint": "^8.15.0",
+    "tsup": "^5.10.1",
+    "typescript": "^4.5.3"
+  },
+  "dependencies": {
+    "@aptos-labs/wallet-adapter-react": "*",
+    "@aptos-labs/wallet-adapter-core": "*",
```
0xmaayan: need to install tailwind on this package 
https://tailwindcss.com/docs/installation

packages/wallet-adapter-tailwind-ui/src/WalletDropdown.tsx
```diff
@@ -0,0 +1,115 @@
+import { WalletReadyState } from '@aptos-labs/wallet-adapter-core';
```
0xmaayan: let's not have this package use `core` as a dependency and instead import and export `WalletReadyState` from the `react` package, same as we do for `WalletName` type.
https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-react/src/useWallet.tsx#L13

I can handle it and publish a new version for `react` package that exports this type.

0xmaayan (changes requested):
Testing it locally and it doesnt seem to load the css correctly.

Also, this packages needs to have tailwind installed and configured.
The main idea is that any app can install this package and uses it regardless of the ui framework the app uses, so this package needs to configure a tailwind ui framework and provide a wallet selector using tailwind-ui design.

image

safaiyeh:
Oh interesting that it's not rendering on your end. Okay let me set up the tailwind config

WEB3Zer0:
Thanks for the review @0xmaayan! Odd that its not rendering on next.js but working locally, ill address the comments.

0xmaayan:
> Thanks for the review @0xmaayan! Odd that its not rendering on next.js but working locally, ill address the comments.

Hi @WEB3Zer0 , I tested it locally with the demo app (it is a nextjs app), I do it with running `pnpm turbo run dev` on the root of this monorepo
https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/CONTRIBUTING.md#develop-locally

packages/wallet-adapter-tailwind-ui/src/WalletDropdown.tsx
0xmaayan: @WEB3Zer0 https://github.com/aptos-labs/aptos-wallet-adapter/pull/36
once approved I can bump the version and you can update and use it

PR #38, opened 2023-01-05
Add fewcha wallet

gregnazario:
I haven't been able to get this to build locally for me, but it should add Fewcha wallet to the test dapp, and the list of supported wallets


0xmaayan:
the code looks good and should work. 
I had no problems running it locally, could you make sure you followed these exact steps ?
https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-core/src/types.ts#L26-L42

0xmaayan (approved):
I updated the PR with the fewcha package dependency and tested it locally, seems to work good.

PR #40, opened 2023-01-06
Create wallet selector with ant design package

0xmaayan:
A new package for users to use Wallet Connect button with Wallet Selector modal that works with the adapter react provider package.

- Users can customize button + modal colors (text, background), font size, padding, margin, etc (provides classes for them to override)
- Connect button that 
    - if wallet is connected, display the connected account address and clicking on it disconnects the wallet
    - if wallet is not connected, open up the modal
- Modal displays whether the wallet installed (if it is, option to connect), or needs to install the wallet (redirect to the wallet website/url)

tagging @moonclavedev for design thoughts

image

image


apps/nextjs-example/pages/index.tsx
```diff
@@ -107,7 +108,7 @@ export default function App() {
         <SuccessAlert text={successAlertMessage} />
       )}
       {errorAlertMessage.length > 0 && <ErrorAlert text={errorAlertMessage} />}
-      <h1 className="flex justify-center mt-2 mb-4 text-4xl font-extrabold tracking-tight leading-none text-black">
+      <h1 className=" flex justify-center mt-2 mb-4 text-4xl font-extrabold tracking-tight leading-none text-black">
```
gregnazario: I'm guessing the space doesn't need to be here

gregnazario (approved):

hariria:
Why Ant design?

hardsetting (approved):
Looks good!

0xmaayan:
> Why Ant design?

good question.

1. It is a flexible lightweight UI library that lets us easily customize the components style/design (unlike MUI and friends)
2. It is a known and popular library in the FE community (Hippo used it in their old adapter and Solana provides a antd ui package). Not only in the web3 space but also in web2 space.
3. It supports multiple FE frameworks (react, vue, angular, etc) and doesn't attache us to only one,  and supports server-side and client-side rendering.
4. Some other implementation are in progress (there is an in progress PR to implement tailwindui).  more work needed is on that PR and tailwind doesn't encourage to use it with CRA so need to look at it more.

PR #46, opened 2023-01-10
Add React as a dependency in wallet-adapter-react package

0xmaayan:
For some reason `React` was listed as a `peerDependency` and it can ends up with React package not being installed in `node_modules` - if it is not there, `pnpm` takes the references to whatever it has in the cache, and if it doesn't find it, the package doesnt know where to take it from and you end up with this error

image

This PR adds `react` to the dependencies array (where it should be).

Before
Screen Shot 2023-01-10 at 4 18 28 PM

After
Screen Shot 2023-01-10 at 4 18 57 PM

### DIFF
--- a/.changeset/pink-games-smoke.md
+++ b/.changeset/pink-games-smoke.md
@@ -1,0 +1,5 @@
+---
+"@aptos-labs/wallet-adapter-react": patch
+---
+
+Add react package as a dependency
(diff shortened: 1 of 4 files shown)


gregnazario (approved):
Works great!  now it finally works properly locally

PR #51, opened 2023-01-12
revert global tailwind config changes in demo app

0xmaayan:
Previous tailwind config changes were done to prevent collision when using different UI frameworks, this PR fixes it differently so we can still use different UI frameworks in the same app.


banool (approved):
😎

PR #52, opened 2023-01-14
Support Loadable wallet

scottphc:
- Support Loadable wallet
- Add a new optional property `propertyName` for multi-chain wallet which may have different brand name with window property name, such as `window.bloctoAptos`
- Implement multi signature verification

### DIFF
--- a/.changeset/forty-toys-dress.md
+++ b/.changeset/forty-toys-dress.md
@@ -1,0 +1,7 @@
+---
+"@aptos-labs/wallet-adapter-core": patch
+---
+
+Support Lodable wallet
+Implement multi signature verification
+Add a new optional property propertyName for multi-chain wallet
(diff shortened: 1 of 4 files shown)


.changeset/forty-toys-dress.md
```diff
@@ -0,0 +1,5 @@
+---
+"@aptos-labs/wallet-adapter-core": patch
+---
+
+Support Lodable wallet
```
0xmaayan, line 5: could you also add the implementation for signAndVerify change?

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -329,7 +332,40 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
       let verified = false;
       if (Array.isArray(response.signature)) {
         // multi sig wallets
-        // TODO - implement multi sig wallets
+        const { fullMessage, signature, bitmap } = response
```
0xmaayan, line 335: thank you!

0xmaayan (commented):
Thank you for your contribution! 

Regarding the multi-chain wallet support, there is another suggestion on how we can go with it here https://github.com/aptos-labs/aptos-wallet-adapter/issues/57 , curios what are your thoughts

.changeset/forty-toys-dress.md
scottphc: updated

0xmaayan (approved):

PR #65, opened 2023-01-24
Support Blocto wallet

scottphc:
- demo: support Blocto wallet
- ant-design: support Loadable wallet

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203568212096206

### DIFF
--- a/apps/nextjs-example/components/AppContext.tsx
+++ b/apps/nextjs-example/components/AppContext.tsx
@@ -4,9 +4,10 @@ import { MartianWallet } from "@martianwallet/aptos-wallet-adapter";
 import { PontemWallet } from "@pontem/wallet-adapter-plugin";
 import { TrustWallet } from "@trustwallet/aptos-wallet-adapter";
 import { SpikaWallet } from "@spika/aptos-plugin";
-import { AptosWalletAdapterProvider } from "@aptos-labs/wallet-adapter-react";
+import { AptosWalletAdapterProvider, NetworkName } from "@aptos-labs/wallet-adapter-react";
 import { FewchaWallet } from "fewcha-plugin-wallet-adapter";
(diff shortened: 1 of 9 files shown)


apps/nextjs-example/components/AppContext.tsx
```diff
@@ -7,6 +7,9 @@ import { SpikaWallet } from "@spika/aptos-plugin";
 import { AptosWalletAdapterProvider } from "@aptos-labs/wallet-adapter-react";
 import { FewchaWallet } from "fewcha-plugin-wallet-adapter";
 import { MSafeWalletAdapter } from "msafe-plugin-wallet-adapter";
+import { BloctoWallet } from "@blocto/aptos-wallet-adapter-plugin";
+import { NetworkName } from "@aptos-labs/wallet-adapter-core"
```
0xmaayan: let's try keep the `core` package "hidden" from dapps, if we need anything from the package - especially types - we can have the react provider exports it for us. like what we do with `WalletReadyState`
https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/apps/nextjs-example/components/WalletButtons.tsx#L1
https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-react/src/useWallet.tsx#L15

0xmaayan (commented):
Thanks for working on it! 
I left a comment on exporting/importing a type from the `core` package instead of installing that package

apps/nextjs-example/components/AppContext.tsx
scottphc: Updated!

0xmaayan (approved):

PR #68, opened 2023-01-28
[wallet-adapter] add mui design package

jillxuu:
this pr is to create a mui design package for wallet adapter to consolidate wallet connection flow design btw explorer and ans.

https://user-images.githubusercontent.com/121921928/215233554-3ffca236-9401-40d7-b9cf-3f49161ffc42.mov


packages/wallet-adapter-mui-design/dist/index.d.ts
```diff
@@ -0,0 +1,7 @@
+declare type WalletConnectorProps = {
```
0xmaayan: we dont need to checkin `dist` folder.

packages/wallet-adapter-mui-design/package.json
```diff
@@ -0,0 +1,63 @@
+{
+  "name": "@aptos-labs/wallet-adapter-mui-design",
+  "version": "0.1.1",
```
0xmaayan: let's start with `0.1.0`

packages/wallet-adapter-mui-design/src/index.tsx
```diff
@@ -0,0 +1,31 @@
+import { useState } from "react";
```
0xmaayan: nit: this file (`index.tsx`) can only hold an export statement
`export * from "./WalletConnector";`

and create another file, `WalletConnector.tsx` file that holds this current file content.

something simlar to what we have in the antd package
https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-ant-design/src/index.tsx

0xmaayan (changes requested):
Looks like it's coming along well.

Can we also include the nextjs example app changes showing how to use it? when doing it, let's also include a `changeset` file - can sync offline on how to do it

0xmaayan (approved):

PR #81, opened 2023-02-06
feat: add TokenPocket wallet adapter

chendatony31:
Add TokenPocket aptos wallet adapter


0xmaayan:
was looking at the TokenPocket plugin implementation https://www.npmjs.com/package/@tp-lab/aptos-wallet-adapter?activeTab=explore

Are these correct?

1. window key should be `aptos`?
2. `providerName` should be `aptos`?

chendatony31:
Yes, as we are a mobile wallet, so I think there would not be a conflict with other extensions. What's more, I think there should have an interface such as `installed` that can be implemented by the wallets themselves. For example we have `aptos.isTokenPocket` it can perfectly be detected if TokenPocket is installed.

0xmaayan:
I see. The problem with that is that the window `aptos` key used by Petra wallet, so when using TokenPocket it would try to open and connect to Petra on a web app. 

We have a plan to add a better support to mobile wallets and to try and make sense with all the different platforms wallets are using.

For now, I would suggest to use a different key (maybe `tokenPocket`) if it doesn't really matter to you and to make sure there would be no conflicts

chendatony31:
Using `aptos` is by design, as most dapps have integrated with Petra. to be compatible with most dapp we have to use it too. Such as `ethereum` object using by metamask.

We will add other key just like you said maybe `tokenPocket` next month, but we will keep `aptos` either.

0xmaayan:
having all dapps using `aptos` is a bad design and exactly what the adapter solves - support different wallets that use a unique and different namespace (i.e window key). This helps to avoid conflicts and/or collisions (that happened in different networks).

`aptos` key was first used by `Aptos Wallet` which today is `Petra` but in reality shouldn't be used by anyone. Still, `Petra` uses its own namespace https://www.npmjs.com/package/petra-plugin-wallet-adapter?activeTab=explore

chendatony31:
Yes, I agree different wallets should have their own namespace, that is why we are going to add `tokenpocket` as our namespace, but as a dapp developer, including so many wallets' packages is also not a good idea. In particular, the interfaces implemented are basically the same.

I think using wallets' different own namespace is a option. but `aptos` should as a backward option to interact with any other unknown aptos wallets. Just like `ethereum` for EVM chains. To void conflicts they can detect `window.aptos` before injecting their own `aptos`.  That's what most EVM extension wallets have done.

0xmaayan:
@chendatony31 please resolve conflicts

chendatony31:
> @chendatony31 please resolve conflicts

Done

chendatony31:
Hi @0xmaayan, We have change our namespace into `tokenpocket`

0xmaayan (approved):

PR #82, opened 2023-02-06
Add ANS support

### LINKED ISSUE #67: [Feature Request] Add ANS support
# 🚀 Feature Request

We want to add ANS support in the wallet adapter. It means that when account is connected to the dapp, the adapter will fetch its ANS name and add it to the `account` object so the dapp can display it.

## Motivation

In the current version of the adapter, the [AccountInfo](https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-core/src/types.ts#L14) type has account `address` and `publickKey` as mandatory properties.
```
type AccountInfo = {
  address: string;
  publicKey: string | string[];
  minKeysRequired?: number
};
```
This lets dapps the option to display the account address on the app. 
We also want to have the option to display the account ANS name.

## Pitch

To implement this feature we need Wallet to export the current connected `chainId` (so the adapter can read it and query the current ANS endpoint i.e mainnet/testnet) as part of the `network()` function response.

`chainId` is part of the Wallet Standard and currently is an optional prop in the `NetworkInfo` [type](https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-core/src/types.ts#L8)

We want to change it to be a mandatory prop - that would be a breaking change.

Then, we would add an optional ans_name to the  `AccountInfo` type.
```
type AccountInfo = {
  address: string;
  publicKey: string | string[];
  minKeysRequired?: number;
  ansName?: string;
};
```

## Additional context
This is the list of wallets that currently **NOT** returning the `chainId` as part of the `network()` function response
Petra
Martian
TrustWallet
Spika
Fewcha
Blocto
mSafe

0xmaayan:
This PR adds support for ANS names. 
When a user connects a wallet, the adapter fetches the account ANS name by the account address and the current network - ANS supports testnet and mainnet only.
If a name exists for the current account - it would be available for dapps as part of the `AccountInfo` type as `ansName` prop.
If a name is not available - `ansName` prop on `AccountInfo` type would be `null`.

https://user-images.githubusercontent.com/29798064/217095312-9d18fdc4-efcb-4bbd-af99-b57dd692716b.mov

### DIFF
--- a/packages/wallet-adapter-core/src/ans.ts
+++ b/packages/wallet-adapter-core/src/ans.ts
@@ -1,35 +1,19 @@
-import { AptosClient } from "aptos";
-
 const ChainIdToAnsContractAddressMap: Record<string, string> = {
-  "1": "0x867ed1f6bf916171b1de3ee92849b8978b7d1b9e0a8cc982a3d19d535dfd9c0c", // mainnet
-  "2": "0x5f8fd2347449685cf41d4db97926ec3a096eaf381332be4f1318ad4d16a8497c", // testnet
-};
-
-type ANSResponse = {
-  name: string;
+  "1": "mainnet", // mainnet
+  "2": "testnet", // testnet
 };
 
 export const getNameByAddress = async (
   chainId: string,
   address: string
 ): Promise<string | null> => {
-  const client = new AptosClient("https://fullnode.testnet.aptoslabs.com/v1");
-
   try {
-    const ansResource = await client.getAccountResource(
-      ChainIdToAnsContractAddressMap[chainId],
-      `${ChainIdToAnsContractAddressMap[chainId]}::domains::ReverseLookupRegistryV1`
+    if (!ChainIdToAnsContractAddressMap[chainId]) return null;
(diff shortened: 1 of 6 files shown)


packages/wallet-adapter-core/src/ans.ts
```diff
@@ -0,0 +1,21 @@
+const ChainIdToAnsContractAddressMap: Record<string, string> = {
+  "1": "mainnet", // mainnet
+  "2": "testnet", // testnet
```
gregnazario, line 3: I didn't realize testnet worked with ANS still, probably needs an update in the explorer too

gregnazario (approved):
lgtm

packages/wallet-adapter-core/src/ans.ts
0xmaayan: yes. 2 websites
https://www.aptosnames.com/ -> mainnet
https://testnet.aptosnames.com/ -> testnet

Issue #94, opened 2023-02-22
Wallet adapter refresh will lead to couple ms loading where wallet info's not loaded yet


jillxuu:
When refreshing the page, wallet adapter will be refreshed as well and the wallet info takes milliseconds to get loaded. During that period, we see wallet button content change from wallet address to "Connect Wallet" as if wallets were not connected.

Would suggest to have a more graceful way to handle the wallet adapter reloading.

### wallet adapter nextjs example: 

https://user-images.githubusercontent.com/121921928/220491290-5d09d9d3-f137-4420-a5c9-811fb1802d92.mov

### ANS: 

https://user-images.githubusercontent.com/121921928/220491294-2b377c1d-05ab-4c07-987a-1d0115dd76d7.mov

### Explorer: 

https://user-images.githubusercontent.com/121921928/220491297-cfee09b3-aa65-4d5d-9e0d-465e16f6936a.mov

theodao:
<img width="661" alt="image" src="https://github.com/aptos-labs/aptos-wallet-adapter/assets/40337350/132a5062-5a26-4624-9bd0-78b203727278">

When I refresh the page, I try to use the `isLoading` variable but as you can see in the photo, it is not sync with address data received. `isLoading` is false when `address` is `null`. 
```
    try {
      setIsLoading(true);
      walletCore.connect(walletName);
    } catch (e) {
      console.log("connect error", e);
    } finally {
      setIsLoading(false);
    }
```
As I see in the source code, `isLoading` variable set to false before the function `walletCore.connect(walletName);` finish right? @0xmaayan

0xmaayan:
@theodao interesting, it might be because we need to `await` here https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-react/src/WalletProvider.tsx#L56

0xmaayan:
@theodao this should fix it https://github.com/aptos-labs/aptos-wallet-adapter/pull/151

PR #99, opened 2023-03-01
Add isLoading state when wallet tries to connect

0xmaayan:
Following https://github.com/aptos-labs/aptos-wallet-adapter/issues/94 it can take a Wallet a couple of ms to connect, during that time we might want to handle the UI accordingly.

This PR adds `isLoading` state to react provider.
can use it like that
```
const { isLoading } = useWallet();
```

### DIFF
--- a/packages/wallet-adapter-react/src/WalletProvider.tsx
+++ b/packages/wallet-adapter-react/src/WalletProvider.tsx
@@ -45,14 +45,19 @@ export const AptosWalletAdapterProvider: FC<AptosWalletProviderProps> = ({
   const [{ connected, account, network, wallet }, setState] =
     useState(initialState);
 
+  const [isLoading, setIsLoading] = useState<boolean>(true);
+
   const walletCore = useMemo(() => new WalletCore(plugins), []);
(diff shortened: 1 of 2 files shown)


packages/wallet-adapter-react/src/WalletProvider.tsx
```diff
@@ -45,14 +45,19 @@ export const AptosWalletAdapterProvider: FC<AptosWalletProviderProps> = ({
...
+
   const walletCore = useMemo(() => new WalletCore(plugins), []);
   const [wallets, setWallets] = useState<Wallet[]>(walletCore.wallets);
 
   const connect = (walletName: WalletName) => {
     try {
+      setIsLoading(true);
       walletCore.connect(walletName);
     } catch (e) {
       console.log("connect error", e);
```
banool, line 58: Probably not here but in a separate PR, could we also expose the error? Something like:
```
const [error, setError] = useState<SomeStructuredErrorType | undefined>(undefined);
```

banool (approved):
Looks good!

jillxuu (approved):
thanks for fixing the issue!

guotie:
nice!

PR #100, opened 2023-03-04
add openblock wallet support

hottomato-c:
Hello there, 
Dev from OpenBlock wallet team,

We have recently published [OpenBlock plugin](https://www.npmjs.com/package/@openblockhq/aptos-wallet-adapter) for aptos-wallet-adapter.

This plugin is an implementation of `AdapterPlugin` from `'@aptos-labs/wallet-adapter-core'` for **OpenBlock** wallet.

We'd like to add some code examples correspondingly in this repo and list our plugin name in the wallet list section of README.md. 

We've tried following the guidance from the CONTIRBUTION.md as mush as possible. 
Please kindly let us know if anything in this PR is not looking right.

Thank you.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ A monorepo modular wallet adapter developed and maintained by Aptos for wallet a
 - [WELLDONE](https://www.npmjs.com/package/@welldone-studio/aptos-wallet-adapter)
 - [Nightly](https://www.npmjs.com/package/@nightlylabs/aptos-wallet-adapter-plugin)
 - [TokenPocket](https://www.npmjs.com/package/@tp-lab/aptos-wallet-adapter)
+- [OpenBlock](https://www.npmjs.com/package/@openblockhq/aptos-wallet-adapter)
 
 #### Develop Locally
 
--- a/apps/nextjs-example/components/AppContext.tsx
+++ b/apps/nextjs-example/components/AppContext.tsx
@@ -11,8 +11,9 @@ import { FewchaWallet } from "fewcha-plugin-wallet-adapter";
 import { MSafeWalletAdapter } from "msafe-plugin-wallet-adapter";
 import { BloctoWallet } from "@blocto/aptos-wallet-adapter-plugin";
 import { WelldoneWallet } from "@welldone-studio/aptos-wallet-adapter";
-import { NightlyWallet } from '@nightlylabs/aptos-wallet-adapter-plugin'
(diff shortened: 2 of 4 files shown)


apps/nextjs-example/pages/index.tsx
```diff
@@ -8,6 +8,8 @@ import dynamic from "next/dynamic";
 import Image from "next/image";
 import { useAutoConnect } from "../components/AutoConnectProvider";
 
+import "regenerator-runtime";
```
0xmaayan: what is this for..?

hottomato-c: hey ya, forgot to mention in the pr. basically it's adding async runtime support for those browsers that doesn't have async support out of the box.

0xmaayan: I see. why is it needed here...? is this needed to support your wallet?

hottomato-c: yep. our plugin currently use async func for simplicity... 
thought maybe we can add some browser support for those users who use earlier version of browsers, just in case...

hottomato-c: and it def can be extracted into another separate pr. let me know if that's the way you prefer

0xmaayan: yes lets remove it from this PR. 

The adapter itself uses async functions and we haven't had any issues with that, so dont think it is needed.

hottomato-c: ✅ just removed that pkg. should be cleaner now.

0xmaayan (approved):

0xmaayan:
@nikli2009 I approved and merged, but https://aptos-labs.github.io/aptos-wallet-adapter/ outputs a console error `regeneratorRuntime is not defined` - that breaks `openblock` support. You might want to add it to the openblock npm plugin [package](https://www.npmjs.com/package/@openblockhq/aptos-wallet-adapter?activeTab=readme)...?

hottomato-c:
> @nikli2009 I approved and merged, but https://aptos-labs.github.io/aptos-wallet-adapter/ outputs a console error `regeneratorRuntime is not defined` - that breaks `openblock` support. You might want to add it to the openblock npm plugin [package](https://www.npmjs.com/package/@openblockhq/aptos-wallet-adapter?activeTab=readme)...?

🤔 that's a good idea, I guess will have to submit another pr to pump the version of the plugin later.

PR #105, opened 2023-03-17
[wallets] Alphabetize wallets

gregnazario:
The wallets were kinda in whatever random order they were added.  This reorders them to be alphabetical


0xmaayan (approved):

0xmaayan (commented):
Usually we use changesets to manage the changelog file. Not sure what would happen if modifying it manually

PR #120, opened 2023-04-13
update package versions

0xmaayan:
update to release new versions b/c https://github.com/aptos-labs/aptos-wallet-adapter/pull/111 breaks adapter on new installs


gregnazario (approved):
Please update the change log to reflect the revert and bug

PR #129, opened 2023-04-22
[fix] Fix wallet adapter auto reconnect

geekflyer:
This PR fixes the wallet adapter losing its connection on page reload and repeatedly asking the user to reconfirm the connection on both explorer and ANS.

In particular this fixes GP-221 (and partially WAL-341) and makes the workaround introduced in https://github.com/aptos-labs/explorer/pull/476 obsolete. See also https://aptos-org.slack.com/archives/C04EXV6UTR7/p1681963789169939?thread_ts=1681962123.674429&cid=C04EXV6UTR7 .

Previously upon page reload a sequence of events triggered a disconnect:
1. The dApp successfully connected to the wallet via autoconnect (which simply calls `connect(<walletName>` )
2. A readyStateChange event from WalletCore triggered a state change in the `wallets` component state by calling `setWallets` here https://github.com/aptos-labs/aptos-wallet-adapter/pull/129/files#diff-976f6ce28f5861155298dbb0cfae197c7ba5f62c12fb8d6c51f603841b8bbcfdR173-R186 , even though none of the items actually changed (but a new array identity had been constructed).
3. This in turn triggered a 2nd auto-reconnect effect under https://github.com/aptos-labs/aptos-wallet-adapter/pull/129/files#diff-976f6ce28f5861155298dbb0cfae197c7ba5f62c12fb8d6c51f603841b8bbcfdR106-R112  which called `connect(<walletName>)` again.
4. The 2nd `connect(<walletName>`)` disconnected the wallet https://github.com/aptos-labs/aptos-wallet-adapter/pull/129/files#diff-8a76cc616d7961765eb7ff494b4e86bc0e56cce6d0273caaa57af714ac95757eR201 and asked the user to reconfirm the connection in a new connection prompt.

This PR fixes 2 things:
1. Double triggering the auto reconnection effect is avoided by using the items in the `wallet` component state as trigger dependency instead of the array identity. So unless actually any of the items or their order changed this won't trigger a reconnect.
2. When `connect(<walletName>)` is called with the same wallet name that is already connected we simply return early from this function and do nothing.

### DIFF
--- a/packages/wallet-adapter-core/src/WalletCore.ts
+++ b/packages/wallet-adapter-core/src/WalletCore.ts
@@ -195,6 +195,9 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
       }
 
       if (this._connected) {
+        // if the selected wallet is already connected, we don't need to connect again
+        if (this.wallet?.name === walletName) return;
+
         await this.disconnect();
       }
       this._connecting = true;
--- a/packages/wallet-adapter-react/src/WalletProvider.tsx
+++ b/packages/wallet-adapter-react/src/WalletProvider.tsx
@@ -109,14 +109,14 @@ export const AptosWalletAdapterProvider: FC<AptosWalletProviderProps> = ({
         connect(localStorage.getItem("AptosWalletName") as WalletName);
       }
     }
-  }, [wallets]);
+  }, wallets);
 
   useEffect(() => {
     if (connected) {
       walletCore.onAccountChange();
       walletCore.onNetworkChange();
     }
-  }, [wallets, connected]);
+  }, [...wallets, connected]);
 
   // Handle the adapter's connect event
   const handleConnect = () => {
@@ -169,20 +169,11 @@ export const AptosWalletAdapterProvider: FC<AptosWalletProviderProps> = ({
     });
   }, [connected]);
 
-  // Handle the adapter's ready state change event
+  // Whenever the readyState of any supported wallet changes we produce a new wallets state array
+  // which in turn causes consumers of the `useWallet` hook to re-render.
+  // See https://github.com/aptos-labs/aptos-wallet-adapter/pull/129#issuecomment-1519026572 for reasoning.
   const handleReadyStateChange = (wallet: Wallet) => {
-    setWallets((prevWallets) => {
-      const index = prevWallets.findIndex(
-        (currWallet) => currWallet === wallet
-      );
-      if (index === -1) return prevWallets;
-
-      return [
-        ...prevWallets.slice(0, index),
-        wallet,
-        ...prevWallets.slice(index + 1),
-      ];
-    });
+    setWallets((wallets) => [...wallets]);
   };
 
   useEffect(() => {
@@ -198,7 +189,7 @@ export const AptosWalletAdapterProvider: FC<AptosWalletProviderProps> = ({
       walletCore.off("networkChange", handleNetworkChange);
       walletCore.off("readyStateChange", handleReadyStateChange);
     };
-  }, [wallets, connected]);
+  }, [...wallets, connected]);
 
   return (
     <WalletContext.Provider
--- a/.changeset/tender-poets-taste.md
+++ b/.changeset/tender-poets-taste.md
@@ -1,0 +1,6 @@
+---
+"@aptos-labs/wallet-adapter-core": patch
+"@aptos-labs/wallet-adapter-react": patch
+---
+
+Fix wallet adapter auto reconnect on page refresh


0xmaayan:
Thanks for looking into it! could you please make sure you follow our [Contributing guide](https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/CONTRIBUTING.md#creating-a-pull-request) and add those changes using `changesets`.

geekflyer:
> ` [0707b2f](/aptos-labs/aptos-wallet-adapter/pull/129/commits/0707b2f73dc37ef5c2c822954a463330d6f89195) `
> [@geekflyer](/geekflyer) [geekflyer](/geekflyer) [force-pushed](/aptos-labs/aptos-wallet-adapter/compare/e30ef08dd9713aa30bc04ac4064295ba727439d9..0707b2f73dc37ef5c2c822954a463330d6f89195) the fix-wallet-auto-reconnect branch from [ `e30ef08` ](/aptos-labs/aptos-wallet-adapter/commit/e30ef08dd9713aa30bc04ac4064295ba727439d9) to [ `0707b2f` ](/aptos-labs/aptos-wallet-adapter/commit/0707b2f73dc37ef5c2c822954a463330d6f89195) [ Compare ](/aptos-labs/aptos-wallet-adapter/compare/e30ef08dd9713aa30bc04ac4064295ba727439d9..0707b2f73dc37ef5c2c822954a463330d6f89195) [ April 22, 2023 22:23 ](#event-9074229132)
> ## Merge state
> Add more commits by pushing to the `[fix-wallet-auto-reconnect](/aptos-labs/aptos-wallet-adapter/tree/fix-wallet-auto-reconnect)` branch on [aptos-labs/aptos-wallet-adapter](/aptos-labs/aptos-wallet-adapter).

done

geekflyer:
> Thanks for looking into it! could you please make sure you follow our [Contributing guide](https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/CONTRIBUTING.md?rgh-link-date=2023-04-23T05%3A14%3A48Z#creating-a-pull-request) and add those changes using `changesets`.

done

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -195,6 +195,9 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
       }
 
       if (this._connected) {
+        // if the selected wallet is already connected, we don't need to connect again
+        if (this.wallet.name === walletName) return;
```
0xmaayan: lets return a message saying `${walletName} is already connected` or something like that, so dapps can let their users know why nothing is happening

0xmaayan (approved):

packages/wallet-adapter-core/src/WalletCore.ts
geekflyer: return or log this?

0xmaayan: return, so wallets can display it to the end user however they want. something like we do [here](https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-core/src/WalletCore.ts#L406)

geekflyer: ok, I added some result structure and enum

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -180,7 +190,7 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
   @emit emits "connect" event
   @throws WalletConnectionError
   */
-  async connect(walletName: WalletName): Promise<void> {
+  async connect(walletName: WalletName): Promise<WalletConnectResult> {
```
0xmaayan: lets not change it, this would actually introduce a breaking change for both dapps and wallet plugins.

Let's actually keep this PR simple and to unblock simply just `return`, we will then design it better

0xmaayan (changes requested):

packages/wallet-adapter-core/src/WalletCore.ts
geekflyer: why is this a breaking change? I assume existing wallet adapter consumers don't do anything with the connect call result at the moment because it always returns void. So if we now return something they will still ignore the result.

0xmaayan: changing the return type creates a breaking change

geekflyer: ok, reverted to the previous PR state.

geekflyer:
btw, what's the actual intention behind this logic? https://github.com/aptos-labs/aptos-wallet-adapter/blob/30caf1ffa74deb16ef8dff60ce305411fe84b1bb/packages/wallet-adapter-react/src/WalletProvider.tsx#L174-L184

We can simplify this code actually to:

```ts
    setWallets((prevWallets) =>
      prevWallets.includes(wallet) ? [...prevWallets] : prevWallets
    );
```
but I still don't quit understand the intention.

0xmaayan (approved):

0xmaayan:
> btw, what's the actual intention behind this logic?
> 
> https://github.com/aptos-labs/aptos-wallet-adapter/blob/30caf1ffa74deb16ef8dff60ce305411fe84b1bb/packages/wallet-adapter-react/src/WalletProvider.tsx#L174-L184
> 
> We can simplify this code actually to:
> 
> ```ts
>     setWallets((prevWallets) =>
>       prevWallets.includes(wallet) ? [...prevWallets] : prevWallets
>     );
> ```
> 
> but I still don't quit understand the intention.

the idea is to update the `wallet.readyState` and broadcast it to the dapp. We assign the `readyState` value here https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-core/src/WalletCore.ts#L67, update the state in the snippet you shared and then dapps can use it like this https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-mui-design/src/WalletModel.tsx#L47

0xmaayan:
fyi, https://github.com/aptos-labs/aptos-wallet-adapter/issues/130

packages/wallet-adapter-react/src/WalletProvider.tsx
```diff
@@ -39,16 +40,18 @@ const initialState: {
 
 export const AptosWalletAdapterProvider: FC<AptosWalletProviderProps> = ({
   children,
-  plugins,
+  plugins: wallets,
```
0xmaayan: Why this change?

packages/wallet-adapter-react/src/WalletProvider.tsx
```diff
@@ -169,21 +172,9 @@ export const AptosWalletAdapterProvider: FC<AptosWalletProviderProps> = ({
...
-      return [
-        ...prevWallets.slice(0, index),
-        wallet,
-        ...prevWallets.slice(index + 1),
-      ];
-    });
-  };
+  // Whenever the readyState of any supported wallet changes which we force a rerender.
+  // See https://github.com/aptos-labs/aptos-wallet-adapter/pull/129#issuecomment-1519026572 for reasoning.
+  const handleReadyStateChange = (wallet: Wallet) => forceUpdate();
```
0xmaayan: can we get this change in a separate PR and keep this one focused on fixing the autoconnect?

packages/wallet-adapter-react/src/WalletProvider.tsx
```diff
@@ -169,20 +169,11 @@ export const AptosWalletAdapterProvider: FC<AptosWalletProviderProps> = ({
...
-      );
-      if (index === -1) return prevWallets;
-
-      return [
-        ...prevWallets.slice(0, index),
-        wallet,
-        ...prevWallets.slice(index + 1),
-      ];
-    });
+    setWallets((wallets) => [...wallets]);
```
0xmaayan, line 176: I think you produce the same array and not really includes the new updated wallet coming from the event. Lets work on this change in a separate PR

geekflyer: The "updated" wallet is already part of the previous array state. We never create "new" wallets in WalletCore - we just watch for state updates/mutations on them and then call this event handler.
Really what this function does it produce a new array with the same items as before, which is the same thing the previous code did. This refactor just makes this a bit more clear.

geekflyer: Just for more clarity:

The previous code was looking for the "updated" wallet in the previous array by using a reference comparison (i.e. it's looking for the same object):
```
const index = prevWallets.findIndex(
        (currWallet) => currWallet === wallet
      );
```

and then it produced a new array with the same object again at the same position:

```
 return [
        ...prevWallets.slice(0, index),
        wallet,
        ...prevWallets.slice(index + 1),
      ];
```

This basically produced a new array with _identical_ items to the previous one, so we might as well just do `prev => [...prev]` to achieve the same goal with less code.

0xmaayan: I think it updates the array with the wallet object argument that comes from the fired event

geekflyer: it does, but the wallet argument coming from the fired event is the same object reference as the object that is already in the previous array.
Otherwise
```
prevWallets.findIndex(
        (currWallet) => currWallet === wallet
      );
```
would never find a matching item in the previous array and always return `-1`.

```
currWallet === wallet
```
compares the object reference and only returns true if the left and right hand of the operator are the same object reference.

0xmaayan: Does it also do deep comparison? The new wallet argument should include the updated readyState value

geekflyer: No, === does not do a deep comparison. It does a reference equality check checking whether `objA === objB` are indeed the same object.
image

geekflyer: And WalletCore only sets/mutates the `.readyState` property on the existing Wallet object (that is already in the state array) and then re-emits that wallet. It never creates a different "Wallet" with a new `.readyState` property.
https://github.com/aptos-labs/aptos-wallet-adapter/blob/5afd0accc09fb67e736b80b7693ea49cac9c2088/packages/wallet-adapter-core/src/WalletCore.ts#L58-L61

packages/wallet-adapter-core/src/WalletCore.ts
banool: I'm not sure how this is a breaking change still, what Christian says makes sense to me?

0xmaayan: it is an optional breaking change from the [wallet plugin side](https://github.com/aptos-labs/wallet-adapter-plugin-template/blob/main/src/index.ts#L26), that can impact dapps. looks like the [interface](https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-core/src/types.ts#L56) for this function returns `any`, so it might be ok.

Issue #98, opened 2023-03-01
[Wallet Adapter] Add deep-link to mobile app support


0xmaayan:
## Motivation
Different wallets support different platforms (desktop, mobile, etc). Our adapter currently supports only web extension wallets and in-app browser on mobile wallet. 

If user interact with the adapter on a web app, the connect button would try to connect the wallet as a web extension wallet. If they try to connect on a web app on mobile - nothing happens.

We need to provide support to mobile wallets that when trying to connect them, the adapter would redirect the user to the mobile app browser or the App store. (same as opensea.io is doing)

## Pitch
A wallet plugin would have an optional `deep_link_property`. If wallet support a mobile app, it can add a deep link value (for example, `https://aptos.app/link`).
when user goes to a dapp that uses the adapter, the adapter detects if it is a mobile or a desktop device.
if it is mobile, the connect button would use the `deep_link_property` (if wallet doesnt have deep_link_property mark as "desktop only")

The adapter will add a `url` query param to the `deep link` that would hold the current website the user in on.
For example - `https://aptos.app/link?url=aptosnames.com`

https://docs.google.com/spreadsheets/d/1MsFNvRNl1fMphrP-bW5MR-8Qd_OQphLOWQKz6K7A5io/edit#gid=0

PR #135, opened 2023-05-11
Deep link

0xmaayan:
Support wallet deep-link https://github.com/aptos-labs/aptos-wallet-adapter/issues/98

Main idea
1. If we are on a desktop view, wallet connect should work as it is now for any wallet type (extension,sdk,etc)
2. If we are on a mobile browser, adapter checks whether a wallet has a `deeplinkProvider` property
  a. If it does, on connect it should redirect the user  to the app (in-app browser) by using the wallet's depp link url
  b. If it does not, up to the dapp to choose on the UI, but can simply disable the button
3. If we are already in a in-app browser, we dont want to redirect anywhere, so connect should work as expected in the mobile app.

### DIFF
--- a/packages/wallet-adapter-core/src/WalletCore.ts
+++ b/packages/wallet-adapter-core/src/WalletCore.ts
@@ -34,6 +34,7 @@ import {
   removeLocalStorage,
   setLocalStorage,
   scopePollingDetectionStrategy,
+  isMobile,
 } from "./utils";
 import { getNameByAddress } from "./ans";
 
@@ -199,6 +200,15 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
         if (this.wallet?.name === walletName) return;
 
         await this.disconnect();
+      }
+      if (isMobile()) {
+        // use wallet deep link
+        if (selectedWallet.deepLink) {
+          const cuurentUrl = window.location.href;
+          window.location.href = `${selectedWallet.deepLink}?url=${cuurentUrl}`;
+        } else {
+          return;
+        }
       }
       this._connecting = true;
       this.setWallet(selectedWallet);
--- a/packages/wallet-adapter-core/src/WalletCore.ts
+++ b/packages/wallet-adapter-core/src/WalletCore.ts
@@ -34,7 +34,7 @@ import {
   removeLocalStorage,
   setLocalStorage,
   scopePollingDetectionStrategy,
-  isMobile,
+  isdRedirectable,
 } from "./utils";
 import { getNameByAddress } from "./ans";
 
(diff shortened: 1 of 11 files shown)


packages/wallet-adapter-core/src/utils/helpers.ts
```diff
@@ -0,0 +1,26 @@
...
+  );
+
+  const isAndroid = /(Android).*Version\/[\d.]+.*Chrome\/[^\s]+ Mobile/i.test(
+    navigator.userAgent
+  );
+
+  return isIphone || isAndroid;
+}
+
+export function isdRedirectable(): boolean {
```
kent-white: dRedirectable?

packages/wallet-adapter-core/src/types.ts
```diff
@@ -52,6 +52,7 @@ export interface AdapterPluginProps<Name extends string = string> {
   url: string;
   icon: `data:image/${"svg+xml" | "webp" | "png" | "gif"};base64,${string}`;
   providerName?: string;
+  deepLink?: string;
```
kent-white: nit: deeplink

one word in this context I think. idk, its debatable

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -200,6 +201,15 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
         await this.disconnect();
       }
+      if (isdRedirectable()) {
+        // use wallet deep link
+        if (selectedWallet.deepLink) {
+          const url = encodeURIComponent(window.location.href);
+          window.location.href = `${selectedWallet.deepLink}?url=${url}`;
+        } else {
+          return;
+        }
```
kent-white: we actually want more than this. We likely want the deeplink parameter to be something like

`deeplinkForUrl: (url: string) -> string`

The reason for this is the wallet may want to link all the way into the in app browser.


`petra://explore?url=google.com`

For people that don't want to return that, they can just return the base link always. `petra://`

kent-white (changes requested):

packages/wallet-adapter-core/src/WalletCore.ts
0xmaayan: interesting. the initial thought was for a wallet to have a `deeplink` prop on its interface where it can be `deeplink=petra://explore` or `deeplink=petra://` and the adapter would concatenate the url to this string.

packages/wallet-adapter-core/src/utils/helpers.ts
```diff
@@ -0,0 +1,26 @@
+export function isMobile(): boolean {
+  return /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(
```
gregnazario: FYI: this might be good too https://gist.github.com/dalethedeveloper/1503252#the-98-solution, I'm amazed no one really has made a standard solution for this

gregnazario (commented):
I defer to @kent-white as I'm not a mobile / TS developer

packages/wallet-adapter-core/src/types.ts
```diff
@@ -53,6 +53,7 @@ export interface AdapterPluginProps<Name extends string = string> {
   icon: `data:image/${"svg+xml" | "webp" | "png" | "gif"};base64,${string}`;
   providerName?: string;
   provider: any;
+  deeplinkProvider?: (data: { url: string }) => string;
```
0xmaayan, line 56: @kent-white

kent-white: beautful

kent-white (approved):

PR #138, opened 2023-05-17
Support deeplink on React Provider and Nextjs demo app

0xmaayan:
adapter core supports deeplink https://github.com/aptos-labs/aptos-wallet-adapter/pull/135 
this PR adds support for the React provider and shows how to use it on the demo app so dapps can use it.

### DIFF
--- a/apps/nextjs-example/components/WalletButtons.tsx
+++ b/apps/nextjs-example/components/WalletButtons.tsx
@@ -1,22 +1,54 @@
-import { useWallet, WalletReadyState } from "@aptos-labs/wallet-adapter-react";
+import {
+  useWallet,
+  WalletReadyState,
+  Wallet,
+  isRedirectable,
+} from "@aptos-labs/wallet-adapter-react";
 
 const WalletButtons = () => {
   const { wallets, connect } = useWallet();
 
   return (
     <>
-      {wallets.map((wallet) => {
-        const isWalletReady = wallet.readyState === WalletReadyState.Installed || wallet.readyState === WalletReadyState.Loadable
-        return (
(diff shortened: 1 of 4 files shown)


apps/nextjs-example/components/WalletButtons.tsx
```diff
@@ -1,22 +1,54 @@
...
+        return !isWalletReady &&
+          isRedirectable() &&
+          !wallet.deeplinkProvider ? (
           <button
-            className={`bg-blue-500  text-white font-bold py-2 px-4 rounded mr-4 ${isWalletReady ? "hover:bg-blue-700" : "opacity-50 cursor-not-allowed"}`}
+            className={`bg-blue-500 text-white font-bold py-2 px-4 rounded mr-4 opacity-50 cursor-not-allowed`}
+            disabled={true}
+            key={wallet.name}
+          >
+            <>{wallet.name} - Desktop Only</>
```
0xmigo, line 35: Curious, why is this "Desktop Only"? Is the condition from line 27-29 only valid for desktop?

0xmaayan: it is for when we are on mobile browser. if the wallet doesn't have `deeplinkProvider` property, the adapter understands that this wallet doesn't support mobile browser. web extension wallets only work on desktop browser

0xmigo: Ah I see!

0xmigo (approved):
LGTM!

nit: Would be great if there is screenshot attached to describe the behavior :)

jillxuu (approved):
thx for adding it!

PR #141, opened 2023-05-24
improvement: allow WalletCore and WalletProvider (AptosWalletAdapterProvider) to receive read-only array of wallets

dmshvetsov:
When an interface receives a non readonly array it tells the user of the interface that this array is intended to be modified which is not the case for `WalletCore`, `WalletProvider` and `useWallet`.

This PR changes `WalletCore` constructor, `useWallet` return and `AptosWalletAdapterProviderProps` types. This is mostly backward compatible change because non read-only `Array` type can be used when `ReadonlyArray` is required but `get wallets(): ReadonlyArray<Wallet>` change is not backward compatible if WalletCore use returned type to modify wallet array. So this might be considered as a major/breaking change because external users of the package will be affected if so maybe we need to bump a major version for `WalletCore`.

### DIFF
--- a/.changeset/wild-peas-own.md
+++ b/.changeset/wild-peas-own.md
@@ -1,0 +1,6 @@
+---
+"@aptos-labs/wallet-adapter-core": minor
+"@aptos-labs/wallet-adapter-react": minor
+---
+
+Support ReadonlyArray of Wallets in AptosWalletAdapterProvider and WalletCore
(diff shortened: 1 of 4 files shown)


0xmaayan:
@dmshvetsov thanks for the PR! the wallet array is not intended to be modified, approving and running tests.

0xmaayan (approved):

PR #143, opened 2023-05-25
run `build` action only on pull request

0xmaayan:
we have `build` and `release` actions running whenever we merge PRs to `main` branch, we should run the `build` action only when there is a new PR and the `release` action when we merge a PR to `main`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,12 +1,9 @@
 name: Build
 
 on:
-  push:
-    branches:
-      - main
   pull_request:
 
-concurrency: 
+concurrency:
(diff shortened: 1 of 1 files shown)


0xmigo:
I think we should do a build whenever there is a new push. New PR + new commits?

0xmaayan:
> I think we should do a build whenever there is a new push. New PR + new commits?

not sure what you mean by `build`. we have 2 jobs - `build` and `release`. the only difference is that `release` also checks changelog for a new release. I don't think we should run that job whenever we push a PR, but run the `build` workflow and on merge, run the `release` workflow.

gregnazario (approved):

banool (approved):

PR #145, opened 2023-06-08
update pontem wallet plugin and fix mobile wallet support on the ui

0xmaayan:


gregnazario (approved):
Should we check for any other wallets that need to be updated too?

PR #151, opened 2023-06-22
Await for wallet connect request before setting `isLoading` state

0xmaayan:
fix for https://github.com/aptos-labs/aptos-wallet-adapter/issues/94#issuecomment-1594252802 

image


banool (approved):
Nice call, good idea to do it this way.

gregnazario (approved):
I'm pretty sure this is exactly why my react demos don't load on refresh

PR #153, opened 2023-06-22
throw `wallet already connected` error and show it on UI

### LINKED ISSUE #130: Return `${walletName} is already connected` message on `connect` method
In this [PR](https://github.com/aptos-labs/aptos-wallet-adapter/pull/129) we check to see whether the selected wallet is already connected and simply `return` if it does. We should return a message indicates that the wallet is already connected so users would understand why nothing is happening when they click the `connect` button.

0xmaayan:
to address https://github.com/aptos-labs/aptos-wallet-adapter/issues/130

https://github.com/aptos-labs/aptos-wallet-adapter/assets/29798064/c54302ed-e98b-4c15-afd1-6d13a85123a7

### DIFF
--- a/apps/nextjs-example/components/WalletButtons.tsx
+++ b/apps/nextjs-example/components/WalletButtons.tsx
@@ -3,26 +3,44 @@ import {
   WalletReadyState,
   Wallet,
   isRedirectable,
+  WalletName,
 } from "@aptos-labs/wallet-adapter-react";
+import { Dispatch, SetStateAction } from "react";
 
-const WalletButtons = () => {
+interface WalletButtonsProps {
+  setErrorAlertMessage: Dispatch<SetStateAction<string>>;
+}
+
+const WalletButtons = (props: WalletButtonsProps) => {
   const { wallets } = useWallet();
 
   return (
     <>
       {wallets.map((wallet: Wallet) => {
-        return walletView(wallet);
+        return walletView(wallet, props.setErrorAlertMessage);
       })}
     </>
   );
 };
 
-const walletView = (wallet: Wallet) => {
+const walletView = (
+  wallet: Wallet,
+  setErrorAlertMessage: Dispatch<SetStateAction<string>>
+) => {
   const { connect } = useWallet();
   const isWalletReady =
     wallet.readyState === WalletReadyState.Installed ||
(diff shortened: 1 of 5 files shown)


apps/nextjs-example/components/WalletButtons.tsx
```diff
@@ -3,26 +3,44 @@ import {
...
     wallet.readyState === WalletReadyState.Loadable;
   const mobileSupport = wallet.deeplinkProvider;
+
+  const onWalletConnectRequest = async (walletName: WalletName) => {
+    try {
+      await connect(walletName);
+    } catch (error: any) {
+      setErrorAlertMessage(error);
+    }
+  };
```
gregnazario, line 42: If there is a way to differentiate between "already connected" and another case, I'd probably make "already connected" a warning alert rather than an error alert.  Just for a UX perspective

gregnazario (approved):

apps/nextjs-example/components/WalletButtons.tsx
0xmaayan: yea I agree! currently the demo app only has success/error alert. In reality any dapp decides its own UI/UX.

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -191,7 +191,10 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
 
       if (this._connected) {
         // if the selected wallet is already connected, we don't need to connect again
-        if (this.wallet?.name === walletName) return;
+        if (this.wallet?.name === walletName)
+          throw new WalletConnectionError(
+            `${walletName} wallet is already connected`
+          ).message;
```
banool, line 197: What's up with the `.message` part of this? Why not just throw the error?

banool (approved):

packages/wallet-adapter-core/src/WalletCore.ts
0xmaayan: yea... this is to only print the `message` ( i.e `Petra wallet is already connected` ) instead of the whole object message which is `name`+`message` (i.e `WalletConnectionError: Petra wallet is already connected`). It relies on the built-in `Error` class.

PR #160, opened 2023-06-30
[ant design] add an optional prop to handle modal open

jillxuu:
### context
this is to fulfill the requirement from our partner game dev shop in this flow so that client can control modal open: 
game-wallet-support drawio

### details
1. change's backward compatible as its optional
2. test it out locally in below screen recording

https://github.com/aptos-labs/aptos-wallet-adapter/assets/121921928/2503df36-5079-4a83-8911-ba1b04e37b0b
3. bump minor version 
4. decide to pass in boolean instead of actions cuz actions will make ant design code messy

### DIFF
--- a/.changeset/bright-jobs-tap.md
+++ b/.changeset/bright-jobs-tap.md
@@ -1,0 +1,5 @@
+---
+"@aptos-labs/wallet-adapter-ant-design": minor
+---
+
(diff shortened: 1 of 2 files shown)


0xmaayan (approved):
beautiful!

PR #163, opened 2023-07-03
support error response on connect/disconnect functions and modify dem…

0xmaayan:
…o app UI


0xmigo (approved):
LGTM!

nit: Would be nice if we can include an image of an updated UI :D

PR #169, opened 2023-07-06
separate connect and connectWallet for SRP

0xmaayan:
If we try to connect an already connected wallet, the error thrown triggers "clearData" which disconnects the existing wallet. This is obviously not the expected behavior. 

This PR separates the `connect` method to `connect()` where we first make all the checks and validations and if all good we call the `connectWallet()` that actually connects to a wallet and throws any wallet connection error.

### DIFF
--- a/packages/wallet-adapter-core/src/WalletCore.ts
+++ b/packages/wallet-adapter-core/src/WalletCore.ts
@@ -172,51 +172,58 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
     }
   }
 
-  /** 
-  Connects a wallet to the app. If a wallet is already connected,
-  we first disconnect the current connected wallet and then connect the selected wallet.
-  On connect success, we set the current account and the network, and keeping the selected wallet
-  name in LocalStorage to support autoConnect function.
-
-  @param walletName. The wallet name we want to connect as a WalletName type.
-  @emit emits "connect" event
-  @throws WalletConnectionError
-  */
+  /**
+   * We first make sure we can connect a dapp to a wallet.
+   * If all good, we connect the wallet by calling `this.connectWallet`
+   * @param walletName. The wallet name we want to connect.
+   */
   async connect(walletName: WalletName): Promise<void | string> {
+    const selectedWallet = this._wallets?.find(
+      (wallet: Wallet) => wallet.name === walletName
+    );
+
+    if (!selectedWallet) return;
+
+    if (this._connected) {
+      // if the selected wallet is already connected, we don't need to connect again
+      if (selectedWallet.name === walletName)
+        throw new WalletConnectionError(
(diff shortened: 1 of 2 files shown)


gregnazario (commented):
So, this still is for anyone who's using `connect`, it will just automatically work?

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -172,51 +172,58 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
-          ).message;
+    const selectedWallet = this._wallets?.find(
+      (wallet: Wallet) => wallet.name === walletName
+    );
+
+    if (!selectedWallet) return;
+
+    if (this._connected) {
+      // if the selected wallet is already connected, we don't need to connect again
+      if (this.wallet?.name === walletName)
```
0xmigo: Wanted to make sure, Is there any chance that wallet is connected (this._connected) but this.wallet? return `null` or `None`? or shall we include the check `if (this._connected && this.wallet)`?

0xmaayan:
> So, this still is for anyone who's using `connect`, it will just automatically work?

no changes in how the end user uses it, only thing is that now if you try to connect an already connected wallet, the connection remains

packages/wallet-adapter-core/src/WalletCore.ts
0xmaayan: yep good catch, should probably do `if (selectedWallet.name === walletName)`

0xmaayan: done

0xmigo (approved):

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -172,51 +172,58 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
+
+    if (!selectedWallet) return;
+
+    if (this._connected) {
+      // if the selected wallet is already connected, we don't need to connect again
+      if (selectedWallet.name === walletName)
+        throw new WalletConnectionError(
+          `${walletName} wallet is already connected`
+        ).message;
+    }
```
gregnazario, line 193: Note for usability, we may want to make this idempotent (I click on the button and I'm already connected, nothing happens) and just log that it's already connected in a debug level.

gregnazario (approved):

packages/wallet-adapter-core/src/WalletCore.ts
0xmaayan: yep, this is the behavior with this change. We also want the dapp to show an error message to the user on why nothing happens, ofc there are many ways to do it but the current adapter design heavily uses errors, so prob easiest would be to throw here also.

PR #170, opened 2023-07-10
Add optional `onError` to catch errors on actions

BriungRi:
# Description

Currently, `connect` throws an error when something goes wrong. This frequently happens when a user cancels the wallet connection process. When a dApp has their own connect wallet modal, the dApp can catch and handle the error. However, when a dApp is using the given Antd modal from this library, there is no way to control the unhandled errors. This diff adds an optional `onError` prop that when provided, will use the `onError` handle rather than throwing an error.

# Test Plan

Apply this diff and test the error handling in the example app

```
diff --git a/apps/nextjs-example/components/AppContext.tsx b/apps/nextjs-example/components/AppContext.tsx
index a954c00..dfe6ee8 100644
--- a/apps/nextjs-example/components/AppContext.tsx
+++ b/apps/nextjs-example/components/AppContext.tsx
@@ -16,10 +16,7 @@ import {
   NetworkName,
 } from "@aptos-labs/wallet-adapter-react";
 
-import {
-  AutoConnectProvider,
-  useAutoConnect,
-} from "./AutoConnectProvider";
+import { AutoConnectProvider, useAutoConnect } from "./AutoConnectProvider";
 import { FC, ReactNode } from "react";
 import face from "../lib/faceInitialization";
 
@@ -47,7 +44,13 @@ const WalletContextProvider: FC<{ children: ReactNode }> = ({ children }) => {
   ];
 
   return (
-    <AptosWalletAdapterProvider plugins={wallets} autoConnect={autoConnect}>
+    <AptosWalletAdapterProvider
+      plugins={wallets}
+      autoConnect={autoConnect}
+      onError={(err) => {
+        console.error(err);
+      }}
+    >
       {children}
     </AptosWalletAdapterProvider>
   );
@@ -59,4 +62,4 @@ export const AppContext: FC<{ children: ReactNode }> = ({ children }) => {
       <WalletContextProvider>{children}</WalletContextProvider>
     </AutoConnectProvider>
   );
-};
\ No newline at end of file
+};
```

The error is properly caught and logged.

### DIFF
--- a/packages/wallet-adapter-react/src/WalletProvider.tsx
+++ b/packages/wallet-adapter-react/src/WalletProvider.tsx
@@ -1,5 +1,4 @@
 import {
-  ErrorInfo,
   FC,
   ReactNode,
   useCallback,
--- a/packages/wallet-adapter-react/src/WalletProvider.tsx
+++ b/packages/wallet-adapter-react/src/WalletProvider.tsx
@@ -1,4 +1,5 @@
 import {
+  ErrorInfo,
   FC,
   ReactNode,
   useCallback,
@@ -24,6 +25,7 @@ export interface AptosWalletProviderProps {
   children: ReactNode;
   plugins: ReadonlyArray<Wallet>;
   autoConnect?: boolean;
+  onError?: (error: any) => void;
 }
 
 const initialState: {
@@ -42,6 +44,7 @@ export const AptosWalletAdapterProvider: FC<AptosWalletProviderProps> = ({
   children,
   plugins,
   autoConnect = false,
+  onError,
 }: AptosWalletProviderProps) => {
   const [{ connected, account, network, wallet }, setState] =
     useState(initialState);
@@ -61,7 +64,8 @@ export const AptosWalletAdapterProvider: FC<AptosWalletProviderProps> = ({
       await walletCore.connect(walletName);
     } catch (error: any) {
       console.log("connect error", error);
-      throw error;
+      if (onError) onError(error);
+      else throw error;
     } finally {
       setIsLoading(false);
     }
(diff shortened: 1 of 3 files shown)


0xmaayan:
I think it is a very needed option. Would it be possible to add an example of how to use it in the demo app? another thing is that in a case one doesn't use the optional `onError` option, the issue with the current Antd modal still persists, we might want to deal with it also by wrapping the `connect` with `try/catch` and use `await` [here](https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-ant-design/src/WalletSelector.tsx#L41-L47) same as we do [here](https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/apps/nextjs-example/components/WalletButtons.tsx#L37-L41).
Other than that, could you [update the changelog](https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/CONTRIBUTING.md#changesets) ?

packages/wallet-adapter-react/src/WalletProvider.tsx
```diff
@@ -1,4 +1,5 @@
 import {
+  ErrorInfo,
```
jillxuu: is it needed?

BriungRi:
> I think it is a very needed option. Would it be possible to add an example of how to use it in the demo app? another thing is that in a case one doesn't use the optional `onError` option, the issue with the current Antd modal still persists, we might want to deal with it also by wrapping the `connect` with `try/catch` and use `await` [here](https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-ant-design/src/WalletSelector.tsx#L41-L47) same as we do [here](https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/apps/nextjs-example/components/WalletButtons.tsx#L37-L41). Other than that, could you [update the changelog](https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/CONTRIBUTING.md#changesets) ?

I think it is better to not catch the `connect` in the antd modal because then users who use the antd modal cannot handle the error themselves. For example, imagine a dapp wants to use their toast library to generate a toast that the user cancelled the connection

0xmaayan (approved):

PR #171, opened 2023-07-11
Enforce types in AdapterPluginEvents.onNetworkChange

BriungRi:
# Description

Enforce types based on the way `WalletCore` calls `onNetworkChange`. This diff depends on [a PR in petra-plugin-wallet-adapter](https://github.com/aptos-labs/petra-plugin-wallet-adapter/pull/7) for the wallet to conform to the new type.

### DIFF
--- a/packages/wallet-adapter-core/src/types.ts
+++ b/packages/wallet-adapter-core/src/types.ts
@@ -24,7 +24,7 @@ export interface AptosWalletErrorResult {
   message: string;
 }
 
-type OnNetworkChange = (
+export type OnNetworkChange = (
   callBack: (networkInfo: NetworkInfo) => Promise<void>
 ) => Promise<void>;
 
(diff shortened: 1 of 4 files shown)


packages/wallet-adapter-core/src/types.ts
```diff
@@ -24,6 +24,20 @@ export interface AptosWalletErrorResult {
   message: string;
 }
 
+type OnNetworkChange = (
```
0xmaayan:
```suggestion
export type OnNetworkChange = (
```

Also, if we make this change, we should probably also update the wallet plugin template with the new type https://github.com/aptos-labs/wallet-adapter-plugin-template/blob/main/src/index.ts#L145

0xmaayan:
please update changelog so we can publish a new version once this PR lands

packages/wallet-adapter-react/src/WalletProvider.tsx
```diff
@@ -1,4 +1,5 @@
 import {
+  ErrorInfo,
```
0xmaayan: hmm I think those changes are from a different PR?

0xmaayan (approved):

PR #175, opened 2023-07-12
Refactor demo app to use onError adapter provider prop

0xmaayan:
[PR](https://github.com/aptos-labs/aptos-wallet-adapter/pull/170) introduced an `onError` prop in the adapter provider, this PR makes use of this prop and refactor the demo app to use it. We also update the doc with this prop.

https://github.com/aptos-labs/aptos-wallet-adapter/assets/29798064/e19398ad-91f6-4354-a8f4-1c6d9af69faf

### DIFF
--- a/apps/nextjs-example/components/Alert.tsx
+++ b/apps/nextjs-example/components/Alert.tsx
@@ -1,8 +1,18 @@
+import { Dispatch, SetStateAction } from "react";
+
 type Alert = {
   text: string;
 };
 
-export function SuccessAlert({ text }: Alert) {
+type SuccessAlert = Alert & {
+  setSuccessAlertMessage: Dispatch<SetStateAction<string>>;
+};
+
+type ErrorAlert = Alert & {
+  setErrorAlertMessage: Dispatch<SetStateAction<string>>;
+};
+
+export function SuccessAlert({ text, setSuccessAlertMessage }: SuccessAlert) {
   return (
     <div
       className="bg-teal-100 border border-teal-400 text-teal-900 px-4 py-3 rounded relative"
@@ -17,7 +27,10 @@ export function SuccessAlert({ text }: Alert) {
           {text}
         </a>
       </span>
-      <span className="absolute top-0 bottom-0 right-0 px-4 py-3">
+      <span
+        className="absolute top-0 bottom-0 right-0 px-4 py-3"
+        onClick={() => setSuccessAlertMessage("")}
+      >
         <svg
           className="fill-current h-6 w-6 text-teal-500"
           role="button"
@@ -32,14 +45,17 @@ export function SuccessAlert({ text }: Alert) {
   );
 }
 
-export function ErrorAlert({ text }: Alert) {
+export function ErrorAlert({ text, setErrorAlertMessage }: ErrorAlert) {
   return (
     <div
       className="bg-red-100 border border-red-400 text-red-700 px-4 py-3 rounded relative"
       role="alert"
     >
       <span className="block sm:inline break-all">{text}</span>
-      <span className="absolute top-0 bottom-0 right-0 px-4 py-3">
+      <span
+        className="absolute top-0 bottom-0 right-0 px-4 py-3"
+        onClick={() => setErrorAlertMessage("")}
+      >
         <svg
           className="fill-current h-6 w-6 text-red-500"
           role="button"
--- a/apps/nextjs-example/components/Alert.tsx
+++ b/apps/nextjs-example/components/Alert.tsx
@@ -2,17 +2,10 @@ import { Dispatch, SetStateAction } from "react";
 
 type Alert = {
   text: string;
+  setText: Dispatch<SetStateAction<string | null>>;
 };
 
-type SuccessAlert = Alert & {
-  setSuccessAlertMessage: Dispatch<SetStateAction<string>>;
-};
-
-type ErrorAlert = Alert & {
-  setErrorAlertMessage: Dispatch<SetStateAction<string>>;
-};
-
-export function SuccessAlert({ text, setSuccessAlertMessage }: SuccessAlert) {
+export function SuccessAlert({ text, setText }: Alert) {
   return (
     <div
       className="bg-teal-100 border border-teal-400 text-teal-900 px-4 py-3 rounded relative"
@@ -29,7 +22,7 @@ export function SuccessAlert({ text, setSuccessAlertMessage }: SuccessAlert) {
       </span>
       <span
         className="absolute top-0 bottom-0 right-0 px-4 py-3"
-        onClick={() => setSuccessAlertMessage("")}
+        onClick={() => setText(null)}
       >
         <svg
           className="fill-current h-6 w-6 text-teal-500"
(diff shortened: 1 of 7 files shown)


apps/nextjs-example/components/AlertProvider.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+interface AlertContextState {
+  successAlertMessage: string;
+  setSuccessAlertMessage: Dispatch<SetStateAction<string>>;
+  errorAlertMessage: string;
+  setErrorAlertMessage: Dispatch<SetStateAction<string>>;
+}
+
+export const AlertContext = createContext<AlertContextState>(
+  {} as AlertContextState
```
BriungRi: I personally prefer `createContext<AlertContextState | undefined>(undefined)`

and later

```
export function useAlert(): AlertContextState {
  const context = useContext(AlertContext);
  if (!context) throw new Error("useAlert must be used within an AlertProvider")
  return context;
}
```
mainly because `{} as AlertContextState` seems like it could work unexpectedly when a provider isn't present in the DOM

apps/nextjs-example/components/AlertProvider.tsx
```diff
@@ -0,0 +1,55 @@
...
+);
+
+export function useAlert(): AlertContextState {
+  return useContext(AlertContext);
+}
+
+export const AlertProvider: FC<{ children: ReactNode }> = ({ children }) => {
+  const [successAlertMessage, setSuccessAlertMessage] = useState<string>("");
```
BriungRi: I usually find `null` acts as a more distinct empty state than `""`

apps/nextjs-example/components/Alert.tsx
```diff
@@ -1,8 +1,18 @@
+import { Dispatch, SetStateAction } from "react";
+
 type Alert = {
   text: string;
```
BriungRi, line 4: To unify the types, could just do

```
type Alert = {
  text: string;
  resetText: () => void;
}
```

and also consider using `useCallback` in `AlertProvider` to define `resetText`

apps/nextjs-example/components/AppContext.tsx
```diff
@@ -19,9 +19,11 @@ import {
 import { AutoConnectProvider, useAutoConnect } from "./AutoConnectProvider";
 import { FC, ReactNode } from "react";
 import face from "../lib/faceInitialization";
+import { AlertProvider, useAlert } from "./AlertProvider";
 
 const WalletContextProvider: FC<{ children: ReactNode }> = ({ children }) => {
   const { autoConnect } = useAutoConnect();
+  const {setErrorAlertMessage} = useAlert();
```
BriungRi: needs format?

apps/nextjs-example/components/WalletButtons.tsx
```diff
@@ -5,29 +5,23 @@ import {
...
-
-const WalletButtons = (props: WalletButtonsProps) => {
+const WalletButtons = () => {
   const { wallets } = useWallet();
 
   return (
     <>
       {wallets.map((wallet: Wallet) => {
-        return walletView(wallet, props.setErrorAlertMessage);
+        return walletView(wallet);
```
BriungRi, line 16: Is there a reason we're not using JSX here? Also, is `key` needed?

apps/nextjs-example/pages/index.tsx
```diff
@@ -51,15 +49,14 @@ export default function App() {
       type_arguments: ["0x1::aptos_coin::AptosCoin"],
       arguments: [account?.address, 1], // 1 is in Octas
     };
+    const response = await signAndSubmitTransaction(payload);
```
BriungRi, line 52: is there a reason we're moving the `signAndSubmitTransaction` outside of the try/catch block? If anything, I would expect `signAndSubmitTransaction` to be relatively prone to throwing errors

apps/nextjs-example/pages/index.tsx
```diff
@@ -41,8 +40,7 @@ export default function App() {
   } = useWallet();
 
   const { autoConnect, setAutoConnect } = useAutoConnect();
-  const [successAlertMessage, setSuccessAlertMessage] = useState<string>("");
-  const [errorAlertMessage, setErrorAlertMessage] = useState<string>("");
```
BriungRi, line 45: can't we still use `setErrorAlertMessage` in this file?

packages/wallet-adapter-react/README.md
```diff
@@ -20,11 +20,12 @@ signAndSubmitTransaction
 signMessage
 ```
 
-##### Feature functions
+##### Feature functions - functions that not all wallets must support
```
BriungRi: `functions that all wallets must support`

or

`functions that may not be supported by all wallets`?

packages/wallet-adapter-react/README.md
```diff
@@ -108,7 +116,11 @@ You can find it [here](../wallet-adapter-ant-design/) with instructions on how t
 ##### connect(walletName)
 
 ```js
-<button onClick={() => connect(wallet.name)}>{wallet.name}</button>
+const onConnect = async (walletName) => {
```
BriungRi, line 119: Shouldn't this be a function that returns a function?

```
const onConnect = (walletName) => {
  return () => connect(walletName);
}
```

BriungRi (commented):
Left comments

apps/nextjs-example/pages/index.tsx
0xmaayan: any downstream error are caught and handled in `onError`

apps/nextjs-example/pages/index.tsx
0xmaayan: we use the global state

packages/wallet-adapter-react/README.md
0xmaayan: it is onClick event that calls a function that calls the `connect` function, it doesnt need to return a function but to call it.

BriungRi: Doesn't `onClick` take in a function?

So in `onClick={onConnect(wallet.name)}`, wouldn't `onConnect(wallet.name)` need to evaluate to a function?

Screenshot 2023-07-13 at 1 00 04 PM

0xmaayan: yes you are right, missed this part. fixed it
```
onClick={() => onConnect(wallet.name)}
```

apps/nextjs-example/pages/index.tsx
```diff
@@ -51,15 +49,14 @@ export default function App() {
...
+    const response = await signAndSubmitTransaction(payload);
     try {
-      const response = await signAndSubmitTransaction(payload);
       await aptosClient.waitForTransaction(response?.hash || "");
       setSuccessAlertMessage(
         `https://explorer.aptoslabs.com/txn/${response?.hash}`
       );
-    } catch (error: any) {
-      console.log("error", error);
-      setErrorAlertMessage(error);
```
BriungRi, line 62: Intentional to not `setErrorAlertMessage` anymore?

apps/nextjs-example/pages/index.tsx
```diff
@@ -51,15 +49,14 @@ export default function App() {
       type_arguments: ["0x1::aptos_coin::AptosCoin"],
       arguments: [account?.address, 1], // 1 is in Octas
     };
+    const response = await signAndSubmitTransaction(payload);
     try {
-      const response = await signAndSubmitTransaction(payload);
       await aptosClient.waitForTransaction(response?.hash || "");
       setSuccessAlertMessage(
         `https://explorer.aptoslabs.com/txn/${response?.hash}`
```
BriungRi, line 56: Not an issue with this diff, but would be nicer as

`View transaction: https://explorer.aptoslabs.com/txn/${response?.hash}`

BriungRi (approved):
Few more comments. Thank you for addressing my comments!

PR #177, opened 2023-07-14
Add setSuccessAlertHash

BriungRi:
# Description

The `text` component is default to a `<a></a>` component which doesn't make sense for text like signed payload. 
 
- Use ReactNode instead of string
 - Support explorer links for other networks (even though we still really only support devnet)

# Test

[localhost_3001.webm](https://github.com/aptos-labs/aptos-wallet-adapter/assets/8957844/ba84c547-5a3f-4f63-b799-42f7910abb12)

### DIFF
--- a/apps/nextjs-example/components/AlertProvider.tsx
+++ b/apps/nextjs-example/components/AlertProvider.tsx
@@ -4,16 +4,16 @@ import {
   FC,
   ReactNode,
   SetStateAction,
+  useCallback,
   useContext,
   useState,
 } from "react";
 import { ErrorAlert, SuccessAlert } from "./Alert";
 
 interface AlertContextState {
-  successAlertMessage: string | null;
-  setSuccessAlertMessage: Dispatch<SetStateAction<string | null>>;
-  errorAlertMessage: string | null;
-  setErrorAlertMessage: Dispatch<SetStateAction<string | null>>;
+  setSuccessAlertHash: (hash: string, networkName?: string) => void;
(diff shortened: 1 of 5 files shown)


apps/nextjs-example/components/AlertProvider.tsx
```diff
@@ -4,16 +4,16 @@ import {
...
 } from "react";
 import { ErrorAlert, SuccessAlert } from "./Alert";
 
 interface AlertContextState {
-  successAlertMessage: string | null;
-  setSuccessAlertMessage: Dispatch<SetStateAction<string | null>>;
-  errorAlertMessage: string | null;
-  setErrorAlertMessage: Dispatch<SetStateAction<string | null>>;
+  setSuccessAlertHash: (hash: string, networkName?: string) => void;
+  setSuccessAlertMessage: Dispatch<SetStateAction<ReactNode | null>>;
```
0xmaayan, line 15: the only difference in `setSuccessAlertMessage` and `setSuccessAlertHash` is the message content? can't we have one setSuccess function and that would be responsible on the content we want to display?

BriungRi: one of them adds a nice helper for generating a link from `hash`. I figured it would be used more than once so would be nice as an additional export from `useAlert`

0xmaayan: we can probably develop it better but this is just a demo app so should be good for it

0xmaayan (approved):

PR #178, opened 2023-07-20
Convert errors to JSON, as they might not be strings

gregnazario:
This happened to me while testing the wallet, it ended up getting an object error instead, which is problematic.  Since we don't control upstream errors, we need to normalize them as JSON.

### DIFF
--- a/packages/wallet-adapter-react/src/WalletProvider.tsx
+++ b/packages/wallet-adapter-react/src/WalletProvider.tsx
@@ -62,7 +62,7 @@ export const AptosWalletAdapterProvider: FC<AptosWalletProviderProps> = ({
       setIsLoading(true);
       await walletCore.connect(walletName);
     } catch (error: any) {
-      console.log("connect error", error);
+      console.log(`connect error ${JSON.stringify(error)}`);
       if (onError) onError(error);
       else throw error;
     } finally {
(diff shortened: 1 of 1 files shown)


packages/wallet-adapter-react/src/WalletProvider.tsx
```diff
@@ -62,7 +62,7 @@ export const AptosWalletAdapterProvider: FC<AptosWalletProviderProps> = ({
       setIsLoading(true);
       await walletCore.connect(walletName);
     } catch (error: any) {
-      console.log("connect error", error);
+      console.log(`connect error ${JSON.stringify(error)}`);
```
0xmaayan, line 65: might be b/c we expect the message to be either a string or contains a `message` prop if its an object
https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-core/src/WalletCore.ts#L240

gregnazario: Okay, then we need to make it handle the message prop then or is it supposed to automatically work like this?

0xmaayan: ideally wallets would have a message prop - which I believe all have except Petra https://www.notion.so/aptoslabs/Wallets-Audit-in-the-Adapter-7-3-2023-cc6bfb1316394be79bb01eeea7ab1d21?showMoveTo=true

gregnazario:
Closing, error that I saw was literally the string "[object Object]" and can't be JSONified

PR #186, opened 2023-08-10
Update aptos version, plugin versions, and add checks to the wallet demo page

gregnazario:


apps/nextjs-example/package.json
```diff
@@ -11,25 +11,27 @@
...
     "@rise-wallet/wallet-adapter": "^0.1.2",
     "@tp-lab/aptos-wallet-adapter": "^1.0.1",
     "@trustwallet/aptos-wallet-adapter": "^0.1.6",
-    "@welldone-studio/aptos-wallet-adapter": "^0.1.2",
+    "@welldone-studio/aptos-wallet-adapter": "^0.1.4",
     "antd": "^5.1.2",
-    "aptos": "^1.14.0",
-    "fewcha-plugin-wallet-adapter": "^0.1.2",
+    "aptos": "^1.17.0",
+    "ethers": "^5.7.2",
```
0xmaayan, line 33: why ethers is needed? is it a dependency in a wallet?

apps/nextjs-example/pages/index.tsx
```diff
@@ -20,293 +21,385 @@ const WalletSelectorAntDesign = dynamic(
...
+    }
+}
+
+const DEVNET_CLIENT = new AptosClient(DEVNET_NODE_URL, {
+    WITH_CREDENTIALS: false,
+});
+const TESTNET_CLIENT = new AptosClient(TESTNET_NODE_URL, {
+    WITH_CREDENTIALS: false,
 });
+const MAINNET_CLIENT = new AptosClient(MAINNET_NODE_URL, {
```
0xmaayan: can we use?
```
const DEVNET_CLIENT = new Provider(Network.DEVNET)
const TESTNET_CLIENT = new Provider(Network.TESTNET)
const MAINNET_CLIENT = new Provider(Network.MAINNET)
```

apps/nextjs-example/pages/index.tsx
```diff
@@ -20,293 +21,385 @@ const WalletSelectorAntDesign = dynamic(
...
+          </td>
+          <td className="px-8 py-4 w-3/4">
+              <WalletConnector/>
+          </td>
+      </tr>
+  </>;
+}
+
+function WalletProps(props: {account: AccountInfo | null, network: NetworkInfo | null, wallet: WalletInfo | null}) {
+
```
0xmaayan: nit:
```
const {account, network, wallet} = props;
```
then can use only the param

apps/nextjs-example/pages/index.tsx
```diff
@@ -20,293 +21,385 @@ const WalletSelectorAntDesign = dynamic(
...
+                                type="checkbox"
+                                className="sr-only peer"
+                                checked={autoConnect}
+                                readOnly
+                            />
+                            <div
+                                onClick={() => {
+                                    setAutoConnect(!autoConnect);
+                                }}
+                                className="w-11 h-6 bg-gray-200 rounded-full peer  peer-focus:ring-green-300  peer-checked:after:translate-x-full peer-checked:after:border-white after:content-[''] after:absolute after:top-0.5 after:left-[2px] after:bg-white after:border-gray-300 after:border after:rounded-full after:h-5 after:w-5 after:transition-all peer-checked:bg-green-600"
```
0xmaayan, line 215: sigh, tailwind

apps/nextjs-example/pages/index.tsx
```diff
@@ -20,293 +21,385 @@ const WalletSelectorAntDesign = dynamic(
...
-    setSuccessAlertMessage(
-      JSON.stringify({ onSignMessageAndVerify: response })
-    );
-  };
-
   return (
     <div>
       <h1 className="flex justify-center mt-2 mb-4 text-4xl font-extrabold tracking-tight leading-none text-black">
-        Aptos Wallet Adapter Demo (Devnet)
+        Aptos Wallet Adapter Tester ({network?.name ?? ""})
```
0xmaayan: should we be worried for users accidentally submitting transactions to mainnet?

apps/nextjs-example/package.json
gregnazario: This is required by some wallet, I think it's facewallet.  We could just remove it, since it feels messy that a wallet needs a transitive dependency like this.

apps/nextjs-example/pages/index.tsx
gregnazario: None of these are a big deal, but I will make buttons disabled for Mainnet then.

Some wallets only support testnet, so I'm trying to keep it to allow at least connecting in all modes

apps/nextjs-example/pages/index.tsx
```diff
@@ -1,312 +1,380 @@
...
+}
+
+const DEVNET_CLIENT = new Provider(Network.DEVNET);
+const TESTNET_CLIENT = new Provider(Network.TESTNET);
+const MAINNET_CLIENT = new Provider(Network.MAINNET);
 
-const aptosClient = new AptosClient(DEVNET_NODE_URL, {
-  WITH_CREDENTIALS: false,
-});
+const isSendableNetwork = (connected: boolean, network?: string): boolean => {
```
0xmaayan, line 39: should `network` be required here?

apps/nextjs-example/pages/index.tsx
```diff
@@ -1,312 +1,380 @@
...
+            <Col>
+                <WalletConnector/>
+            </Col>
+        </Row>
+    </>;
+}
+
+// TODO: Verify public key matches account
+function WalletProps(props: { account: AccountInfo | null, network: NetworkInfo | null, wallet: WalletInfo | null }) {
+    const {account, network, wallet} = props;
```
0xmaayan, line 112: does `account` needed here?

apps/nextjs-example/pages/index.tsx
```diff
@@ -1,312 +1,380 @@
...
+        <tr>
+            <Col title={true}>
+                <h3>Wallet Name</h3>
+            </Col>
+            <Col>
+                <b>Icon: </b>
+                {props.wallet && (
+                    <Image
+                        src={wallet?.icon ?? ""}
+                        alt={wallet?.name ?? ""}
```
0xmaayan, line 128: do we check a wallet has both icon and name props?

0xmaayan (approved):

banool (approved):

bitcoinfinancier (approved):

PR #187, opened 2023-08-24
Add initial fee payer / multi-agent along with website tests

gregnazario:
This is a first attempt at fee payer and multi agent transactions, though the model is not ideal.

1. Application creates a transaction (and has to set the expiration time etc. probably longer).
2. Calls `provider.signMultiTransaction()` to sign with a local key (or send to a backend to do similar)
3. Calls `signMultiAgentTransaction` to get the user's signature.
4. Converts the User's signature into an Authenticator (tbd, how should we handle this)
5. Application handles submitting the transaction.

This leaves a few issues:
1. User has no control over where the transaction is sent, though the app does
2. Application must properly handle submission and retrieval, which they basically didn't have to do any of before with the wallet adapter
3. Application has to do a lot more processing around the transaction, vs the simple "signAndSubmitTransaction" with a payload previously in the wallet adapter.
4. Simulation of results is usually not handled as easily as before where the transaction would be simulated and submitted by the wallet
5. Wallets now need to handle these multiple paths

Support via Petra wallet adapter added here:
https://github.com/aptos-labs/petra-plugin-wallet-adapter/pull/10

https://github.com/aptos-labs/aptos-core/issues/9896

### DIFF
--- a/apps/nextjs-example/pages/index.tsx
+++ b/apps/nextjs-example/pages/index.tsx
@@ -368,31 +368,6 @@ function OptionalFunctionality() {
         );
     };
 
-    const onSubmitScript = async () => {
-        const payload = {
-            type: "script_payload",
-            code: {
-                bytecode: DO_NOTHING_SCRIPT
-            },
-            type_arguments: [],
-            arguments: []
-        };
-
-        const response = await signAndSubmitTransaction(payload);
-        try {
-            if (response?.hash === undefined) {
-                throw new Error(`No response given ${response}`)
-            }
-            await aptosClient(network?.name.toLowerCase()).waitForTransaction(response.hash);
-            setSuccessAlertHash(response.hash, network?.name);
-        } catch (error) {
-            console.error(error);
-        }
-        setSuccessAlertMessage(
-            JSON.stringify({signAndSubmitTransaction: response})
-        );
-    };
-
     // TODO: Let's put this into a cookie or local storage so it only happens once
     const createFeePayerAccount = async (client: AptosClient) => {
         // Already exists, so lets not fund it
@@ -459,7 +434,6 @@ function OptionalFunctionality() {
         // Submit it TODO: the wallet possibly should send it instead?
         let response: undefined | Types.PendingTransaction = undefined;
             response = await provider.submitFeePayerTransaction(rawTxn, userAuthenticator, feePayerAuthenticator);
-            console.log(`HASH : ${response?.hash}`)
             if (response?.hash === undefined) {
                 throw new Error(`No response given ${response}`)
             }
@@ -481,8 +455,6 @@ function OptionalFunctionality() {
             <Button color={"blue"} onClick={onSignTransaction} disabled={!sendable} message={"Sign transaction"}/>
             <Button color={"blue"} onClick={onSignAndSubmitBCSTransaction} disabled={!sendable}
                     message={"Sign and submit BCS transaction"}/>
-            <Button color={"blue"} onClick={onSubmitScript} disabled={!sendable}
-                    message={"Sign and submit move script"}/>
             <Button color={"blue"} onClick={onSubmitFeePayer} disabled={!sendable}
                     message={"Sign and submit fee payer"}/>
         </Col>
--- a/apps/nextjs-example/pages/index.tsx
+++ b/apps/nextjs-example/pages/index.tsx
@@ -314,9 +314,6 @@ function OptionalFunctionality() {
         signAndSubmitBCSTransaction,
         signTransaction,
         signMessageAndVerify,
-        prepareFeePayerTransaction,
-        prepareMultiAgentTransaction,
-        signAndSubmitMultiAgentTransaction,
         signAndSubmitFeePayerTransaction
     } = useWallet();
     let sendable = isSendableNetwork(connected, network?.name)
@@ -430,17 +427,30 @@ function OptionalFunctionality() {
 
         const payload: Types.TransactionPayload = {
             type: "entry_function_payload",
-            function: "0x1::coin::transfer",
-            type_arguments: ["0x1::aptos_coin::AptosCoin"],
+            function: "0x1::aptos_account::transfer",
+            type_arguments: [],
             arguments: [account.address, 1], // 1 is in Octas
         };
 
-        let prepResponse = await prepareFeePayerTransaction(payload, feePayerAccount.address().hex(), feePayerAccount.pubKey().hex());
-        if (!prepResponse) {
-            // TODO: Handle error
-            throw new Error("Failed to prepare fee payer transaction");
-        }
-        let prepTxn = prepResponse as TxnBuilderTypes.FeePayerRawTransaction;
+        let chainId = await provider.getChainId();
+        let accountResource = await provider.getAccount(account.address);
+        let now = Date.now() / 1000;
+        let rawTxn = new TxnBuilderTypes.RawTransaction(
+            TxnBuilderTypes.AccountAddress.fromHex(account.address),
+            BigInt(accountResource.sequence_number),
+            new TxnBuilderTypes.TransactionPayloadEntryFunction(
+                new TxnBuilderTypes.EntryFunction(
+                    new TxnBuilderTypes.ModuleId(TxnBuilderTypes.AccountAddress.fromHex("0x1"),
+                        new TxnBuilderTypes.Identifier("aptos_account")),
+                    new TxnBuilderTypes.Identifier("transfer"),
+                    [],
+                    [BCS.bcsToBytes(TxnBuilderTypes.AccountAddress.fromHex(account.address)), BCS.bcsSerializeUint64(1)])),
+            BigInt(1000000),
+            BigInt(100),
+        BigInt(now) + BigInt(60000),
+            new TxnBuilderTypes.ChainId(chainId)
+        );
+        let prepTxn = new TxnBuilderTypes.FeePayerRawTransaction(rawTxn, [], TxnBuilderTypes.AccountAddress.fromHex(feePayerAccount.address().hex()));
 
         // This would normally be provided by the service or a server
         let feePayerSignature = await provider.signMultiTransaction(feePayerAccount, prepTxn);
--- a/apps/nextjs-example/pages/index.tsx
+++ b/apps/nextjs-example/pages/index.tsx
@@ -16,8 +16,9 @@ import Image from "next/image";
 import {useAutoConnect} from "../components/AutoConnectProvider";
 import {useAlert} from "../components/AlertProvider";
 import {AccountInfo, NetworkInfo, WalletInfo} from "@aptos-labs/wallet-adapter-core";
+import {useState} from "react";
 
-let feePayerAccount: AptosAccount | undefined = undefined;
+const FEE_PAYER_ACCOUNT_KEY = "feePayerAccountPrivateKey";
 const DO_NOTHING_SCRIPT = "a11ceb0b0600000001050001000000000102";
 
 const WalletButtons = dynamic(() => import("../components/WalletButtons"), {
@@ -304,6 +305,7 @@ function RequiredFunctionality() {
 }
 
 function OptionalFunctionality() {
+    const [initializedFeePayer, setInitializedFeePayer] = useState<boolean>(false);
     const {setSuccessAlertMessage, setSuccessAlertHash} = useAlert();
     // TODO: pass as props
     const {
@@ -368,28 +370,43 @@ function OptionalFunctionality() {
         );
     };
 
+    const loadFeePayerAccount = (): AptosAccount => {
+        // Check if it exists in local storage
+        let maybeFeePayer = window.localStorage.getItem(FEE_PAYER_ACCOUNT_KEY);
+        if (!maybeFeePayer) {
+            let feePayerAccount = new AptosAccount();
+            window.localStorage.setItem(FEE_PAYER_ACCOUNT_KEY, feePayerAccount.toPrivateKeyObject().privateKeyHex)
+            return feePayerAccount;
+        } else {
+            let key = HexString.ensure(maybeFeePayer).toUint8Array();
+            return new AptosAccount(key);
+        }
+    }
+
     // TODO: Let's put this into a cookie or local storage so it only happens once
-    const createFeePayerAccount = async (client: AptosClient) => {
-        // Already exists, so lets not fund it
-        if (feePayerAccount) {
+    const fundAndSetFeePayer = async (client: AptosClient): Promise<AptosAccount> => {
+        // Check if it exists in local storage
+        let feePayerAccount = loadFeePayerAccount();
+        let feePayerAddress = feePayerAccount.address();
+
+        if (initializedFeePayer) {
             return feePayerAccount;
         }
-
-        let newAccount = new AptosAccount();
-        let feePayerAddress = newAccount.address();
 
         try {
             let coinClient = new CoinClient(client);
             let balance = await coinClient.checkBalance(feePayerAddress);
             if (Number(balance) >= 100000000) {
-                return newAccount;
+                setInitializedFeePayer(true);
+                return feePayerAccount;
             }
         } catch (error: any) {
             // Swallow errors, likely due to account not existing
         }
         await (faucet(network?.name.toLowerCase()).fundAccount(feePayerAddress, 100000000));
 
-        return newAccount;
+        setInitializedFeePayer(true);
+        return feePayerAccount;
     }
 
     const onSubmitFeePayer = async () => {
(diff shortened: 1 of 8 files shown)


apps/nextjs-example/pages/index.tsx
```diff
@@ -340,6 +371,95 @@ function OptionalFunctionality() {
...
+            code: {
+                bytecode: DO_NOTHING_SCRIPT
+            },
+            type_arguments: [],
+            arguments: []
+        };
+
+        const response = await signAndSubmitTransaction(payload);
+        try {
+            if (response?.hash === undefined) {
```
0xmaayan: is this needed? I think if `response.hash` is `undefined` then `waitForTransaction` would throw an error?

apps/nextjs-example/pages/index.tsx
```diff
@@ -1,4 +1,14 @@
-import {AptosClient, BCS, Network, Provider, TxnBuilderTypes, Types} from "aptos";
+import {
+    AptosAccount,
+    AptosClient,
+    BCS,
+    CoinClient,
```
0xmaayan, line 5: nite: can use `FungibleAssetClient`

apps/nextjs-example/pages/index.tsx
```diff
@@ -340,6 +371,95 @@ function OptionalFunctionality() {
...
+        if (!account) {
+            throw new Error("Not connected");
+        }
+        let provider = aptosClient(network?.name.toLowerCase());
+        // Generate an account and fund it
+        let feePayerAccount = await createFeePayerAccount(provider);
+
+        const payload: Types.TransactionPayload = {
+            type: "entry_function_payload",
+            function: "0x1::coin::transfer",
```
0xmaayan: nit
```suggestion
            function: "0x1::aptos_account::transfer",
```
then dont need the `type_arguments`

apps/nextjs-example/pages/index.tsx
```diff
@@ -26,15 +39,28 @@ const aptosClient = (network?: string) => {
...
+        throw new Error("Please use devnet or testnet for testing");
     } else {
         throw new Error(`Unknown network: ${network}`);
     }
 }
 
 const DEVNET_CLIENT = new Provider(Network.DEVNET);
 const TESTNET_CLIENT = new Provider(Network.TESTNET);
-const MAINNET_CLIENT = new Provider(Network.MAINNET);
+const DEVNET_FAUCET = new FaucetClient("https://fullnode.devnet.aptoslabs.com", "https://faucet.devnet.aptoslabs.com");
```
0xmaayan, line 63: sigh

apps/nextjs-example/pages/index.tsx
```diff
@@ -340,6 +371,95 @@ function OptionalFunctionality() {
...
+            return feePayerAccount;
+        }
+
+        let newAccount = new AptosAccount();
+        let feePayerAddress = newAccount.address();
+
+        try {
+            let coinClient = new CoinClient(client);
+            let balance = await coinClient.checkBalance(feePayerAddress);
+            if (Number(balance) > 10000) {
```
0xmaayan: why? to prevent rate limit?

apps/nextjs-example/pages/index.tsx
```diff
@@ -340,6 +371,95 @@ function OptionalFunctionality() {
...
+                return newAccount;
+            }
+        } catch (error: any) {
```
0xmaayan, line 403: if we swallow, why even try/catch?

apps/nextjs-example/pages/index.tsx
```diff
@@ -26,15 +39,28 @@ const aptosClient = (network?: string) => {
...
         return TESTNET_CLIENT;
     } else if (network === NetworkName.Mainnet.toLowerCase()) {
-        return MAINNET_CLIENT;
+        throw new Error("Please use devnet or testnet for testing");
+    } else {
+        throw new Error(`Unknown network: ${network}`);
+    }
+}
+
+const faucet = (network?: string) => {
```
0xmaayan, line 49: might worth to get both provider and faucet into a global state in the future, can clean up some of the code

0xmaayan (commented):
overall lgtm.

apps/nextjs-example/pages/index.tsx
gregnazario: There was some case... where it didn't throw an error afaik, or I had to do this for some reason on typescript.

apps/nextjs-example/pages/index.tsx
gregnazario: Yeah, I was trying to reduce extra calls on a local test

apps/nextjs-example/pages/index.tsx
gregnazario: Is there a way to swallow errors without try catch?

apps/nextjs-example/pages/index.tsx
gregnazario: If so, we should make an example for most dapps so it can be a common pattern

0xmaayan: it is a frontend framework pattern, not specifically to aptos/blockchaim/wallet adapter =) can revisit it later

apps/nextjs-example/pages/index.tsx
```diff
@@ -340,6 +368,109 @@ function OptionalFunctionality() {
...
+        const payload = {
+            type: "script_payload",
+            code: {
+                bytecode: DO_NOTHING_SCRIPT
+            },
+            type_arguments: [],
+            arguments: []
+        };
+
+        const response = await signAndSubmitTransaction(payload);
```
xbtmatt: Wait so the only reason scripts don't work is because wallets don't handle the different `type`? 💀

gregnazario: Yes, there's something weird that all wallets do, and inputs will likely have to be preserialized. I will look more into that later

apps/nextjs-example/pages/index.tsx
```diff
@@ -340,6 +368,109 @@ function OptionalFunctionality() {
...
+        } catch (error) {
+            console.error(error);
+        }
+        setSuccessAlertMessage(
+            JSON.stringify({signAndSubmitTransaction: response})
+        );
+    };
+
+    // TODO: Let's put this into a cookie or local storage so it only happens once
+    const createFeePayerAccount = async (client: AptosClient) => {
```
xbtmatt: for the only happens once comment: maybe something like this..?

```typescript
const [feePayerAccount, setFeePayerAccount] = useState<AptosAccount>();

const fundAndSetFeePayer = async(): AptosAccount => {
  const feePayer = new AptosAccount();
  await faucet(network?.name.toLowerCase()).fundAccount(feePayer.address(), 100_000_000));
  setFeePayerAccount(feePayer);
  return feePayer;
}

// create and fund on component mount
useEffect(() => {
  fundAndSetFeePayer();
}, []);

const onSubmitFeePayer = async() => {
  // ...
  const feePayer = feePayerAccount ?? await fundAndSetFeePayer();
  // ...
};
```

could add coin balance check & debounce if necessary, but I don't think it really matters, since they will likely never spend the balance in gas fees

apps/nextjs-example/pages/index.tsx
xbtmatt: sort of tangential (although it is related bc I run into this all the time, locally in scripts and in dapps)

is there a reason we don't just auto-initialize the faucet in the provider to the same network as long as they're not on mainnet? :p

xbtmatt (commented):
lgtm, the points you mention are true but a lot of it (imo) comes down to how the wallet handles and presents the tx to the user. 

The main issues I see are the complex application logic and simulation stuff, but that's more an issue with the transaction builder patterns and the fee payer model

And then adding multisig should be pretty simple/similar

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -371,7 +371,28 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
   }
 
-  /** 
+  async signMultiAgentTransaction(
+    transaction: TxnBuilderTypes.MultiAgentRawTransaction | TxnBuilderTypes.FeePayerRawTransaction
+  ): Promise<string | null> {
+    if (this._wallet && !("signMultiAgentTransaction" in this._wallet)) {
+      throw new WalletNotSupportedMethod(
+          `Multi agent & Fee payer transactions are not supported by ${this.wallet?.name}`
+      ).message;
```
hardsetting, line 380: Why returning the message and not the error?

apps/nextjs-example/pages/index.tsx
hardsetting: > Wait so the only reason scripts don't work is because wallets don't handle the different type?

Wait what do you mean by this?
Petra supports script payloads right?

apps/nextjs-example/pages/index.tsx
hardsetting: Yeah `signAndSubmitTransaction` should fail if the transaction is not submitted.

hardsetting: And the error displayed on the wallet's side

apps/nextjs-example/pages/index.tsx
```diff
@@ -340,6 +368,109 @@ function OptionalFunctionality() {
...
+            type_arguments: [],
+            arguments: []
+        };
+
+        const response = await signAndSubmitTransaction(payload);
+        try {
+            if (response?.hash === undefined) {
+                throw new Error(`No response given ${response}`)
+            }
+            await aptosClient(network?.name.toLowerCase()).waitForTransaction(response.hash);
```
hardsetting: Mmm ok at least what we're doing on Petra is, we're submitting the transaction and waiting for its result.
If there's an error when submitting, we show it in the prompt, but the prompt stays open.

Maybe we should enforce the same for other wallets?
Otherwise then it becomes the dapp's responsibility to display the transaction error, and we have to figure out passing just the right details back to the dapp.

apps/nextjs-example/pages/index.tsx
hardsetting: The idea is to persist the account across sessions, my vote is using `window.localStorage`

apps/nextjs-example/pages/index.tsx
```diff
@@ -340,6 +368,109 @@ function OptionalFunctionality() {
...
+        };
+
+        // This would normally be provided by the service or a server
+        // We need to increase the default timeout
+        let expiration_timestamp_secs = Math.floor(Date.now() / 1000) + 60000;
+
+        const rawTxn = await provider.generateFeePayerTransaction(account.address, payload, feePayerAccount.address().hex(), [], {expiration_timestamp_secs: expiration_timestamp_secs.toString()})
+
+        // Sign with fee payer
+        const feePayerAuthenticator = await provider.signMultiTransaction(feePayerAccount, rawTxn);
```
hardsetting, line 441: I'm confused about `signMultiTransaction`, looking at the implementation [here](https://github.com/aptos-labs/aptos-core/blob/a2877543f2a2a780919f341bd11169fd60ea522e/ecosystem/typescript/sdk/src/providers/aptos_client.ts#L382) there's nothing that should tie this function to `FeePayer` or `MultiAgent` as `TransactionBuilder.getSigningMessage` accepts all kinds of raw transactions

hardsetting: This is maybe more of a feedback for the Aptos client

hardsetting (approved):
Overall looks great!
Just a few questions

hardsetting:
> User has no control over where the transaction is sent, though the app does

I'm curious, what kind of problems could this create?
The dapp can always fetch the current network from the wallet adapter, but iiuc the fact that the dapp could submit the transaction through a random node could be undesireable to the user?

> Application must properly handle submission and retrieval, which they basically didn't have to do any of before with the wallet adapter

dapps that have to use the `signTransaction` endpoint always had to submit transactions from their side.
I agree that supporting sponsored transactions without having extra logic would be very nice and simplify adoption, but I'm not too worried about the current implementation

> Simulation of results is usually not handled as easily as before where the transaction would be simulated and submitted by the wallet

I added experimental support for simulations on `signTransaction` calls [in this PR](https://github.com/aptos-labs/wallet/pull/1517). Hopefully that can ease the issue

> Wallets now need to handle these multiple paths

Instead of having two separate calls, we can merge them and have overloads.

```typescript
type AnyTransactionPayload = TxnBuilderTypes.TransactionPayload | JsonPayload; 
type AnyRawTransaction = TxnBuilderTypes.RawTransaction
  | TxnBuilderTypes.FeePayerRawTransaction
  | TxnBuilderTypes.MultiAgentRawTransaction;

// Hex-encoded signature
type TransactionSignature = string;

interface WalletSignatureAPI {
  signTransaction(payload: AnyTransactionPayload, options: TransactionOptions): Promise<TransactionSignature>;
  signTransaction(rawTxn: AnyRawTransaction): Promise<TransactionSignature>;
} 
```

apps/nextjs-example/pages/index.tsx
hardsetting: Ok I guess that the returned `AccountAuthenticator` is only used as input for authenticators of a multi-agent or a fee-payer transaction.

hardsetting:
If we want to also allow wallet user's to sign and submit a pre-signed sponsored transaction, we should update the signature as follows:
```typescript
type AnyTransactionPayload = TxnBuilderTypes.TransactionPayload | JsonPayload; 
type AnyRawTransaction = TxnBuilderTypes.RawTransaction
  | TxnBuilderTypes.FeePayerRawTransaction
  | TxnBuilderTypes.MultiAgentRawTransaction;

// Hex-encoded signature
type TransactionSignature = string;

interface WalletSignatureAPI {
  signTransaction(payload: AnyTransactionPayload, options: TransactionOptions): Promise<TransactionSignature>;
  signTransaction(rawTxn: AnyRawTransaction): Promise<TransactionSignature>;
  
  signAndSubmitTransaction(payload: AnyTransactionPayload, options: TransactionOptions): Promise<Types.UserTransaction>;
  signAndSubmitTransaction(rawTxn: TxnBuilderTypes.RawTransaction): Promise<Types.UserTransaction>;
  signAndSubmitTransaction(
    rawTxn: TxnBuilderTypes.FeePayerRawTransaction,
    feePayerAuthenticator: TxnBuilderTypes.AccountAuthenticator,
    secondaryAuthenticators?: TxnBuilderTypes.AccountAuthenticator[],
  ): Promise<Types.UserTransaction>;
  signAndSubmitTransaction(
    rawTxn: TxnBuilderTypes.MultiAgentRawTransaction,
    secondaryAuthenticators?: TxnBuilderTypes.AccountAuthenticator[],
  ): Promise<Types.UserTransaction>;
}
```

apps/nextjs-example/pages/index.tsx
gregnazario: Petra only supports it through BCS, not through non-bcs, but I guess I now realize that makes some sense.

packages/wallet-adapter-core/src/WalletCore.ts
gregnazario: That seems wrong, will look

apps/nextjs-example/pages/index.tsx
gregnazario: This is defensive programming in the event that the response returned is not properly formatted

apps/nextjs-example/pages/index.tsx
gregnazario: I've removed the script functionality for now.

This is up to the wallet, but I think we either way have to use waitForTransaction to deal with eventual consistency (the wallet may use a different endpoint)

packages/wallet-adapter-core/src/WalletCore.ts
gregnazario: This seems to be consistent across all other parts of the wallet adapter.

I suspect we need to fix this across all, so I'm leaving it now to be consistent

apps/nextjs-example/pages/index.tsx
gregnazario: Let me follow up in a quick PR after this for local storage and all.

gregnazario:
> > User has no control over where the transaction is sent, though the app does
> 
> I'm curious, what kind of problems could this create? The dapp can always fetch the current network from the wallet adapter, but iiuc the fact that the dapp could submit the transaction through a random node could be undesireable to the user?
> 
> > Application must properly handle submission and retrieval, which they basically didn't have to do any of before with the wallet adapter
> 
> dapps that have to use the `signTransaction` endpoint always had to submit transactions from their side. I agree that supporting sponsored transactions without having extra logic would be very nice and simplify adoption, but I'm not too worried about the current implementation
> 
> > Simulation of results is usually not handled as easily as before where the transaction would be simulated and submitted by the wallet
> 
> I added experimental support for simulations on `signTransaction` calls [in this PR](https://github.com/aptos-labs/wallet/pull/1517). Hopefully that can ease the issue
> 
> > Wallets now need to handle these multiple paths
> 
> Instead of having two separate calls, we can merge them and have overloads.
> 
> ```ts
> type AnyTransactionPayload = TxnBuilderTypes.TransactionPayload | JsonPayload; 
> type AnyRawTransaction = TxnBuilderTypes.RawTransaction
>   | TxnBuilderTypes.FeePayerRawTransaction
>   | TxnBuilderTypes.MultiAgentRawTransaction;
> 
> // Hex-encoded signature
> type TransactionSignature = string;
> 
> interface WalletSignatureAPI {
>   signTransaction(payload: AnyTransactionPayload, options: TransactionOptions): Promise<TransactionSignature>;
>   signTransaction(rawTxn: AnyRawTransaction): Promise<TransactionSignature>;
> } 
> ```

Overloading is problematic, as it definitely can't be used with javascript then (though I don't know what the compatibility is there)

PR #189, opened 2023-08-31
Submit from wallet with fee payer support

gregnazario:
Followup from previous fee payer PR, which adds support for submitting from a wallet directly.  

I'm still working on testing it with Petra or another wallet directly to ensure it is the right model.

https://github.com/aptos-labs/aptos-core/issues/9896

This has been tested with petra as a wallet plugin here https://github.com/aptos-labs/petra-plugin-wallet-adapter/pull/10

### DIFF
--- a/packages/wallet-adapter-core/src/WalletCore.ts
+++ b/packages/wallet-adapter-core/src/WalletCore.ts
@@ -1,4 +1,11 @@
-import {HexString, TxnBuilderTypes, Types} from "aptos";
+import {
+  AnyRawTransaction,
+  HexString,
+  Network,
+  Provider,
+  TxnBuilderTypes,
+  Types
+} from "aptos";
 import EventEmitter from "eventemitter3";
 import nacl from "tweetnacl";
 import { Buffer } from "buffer";
@@ -29,7 +36,7 @@ import {
   Wallet,
   WalletInfo,
   WalletCoreEvents,
-  TransactionOptions,
+  TransactionOptions, RawTransactionRequest, RawTransactionPrepPayload, SignRawTransactionResponse,
 } from "./types";
 import {
   removeLocalStorage,
@@ -371,27 +378,6 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
     }
   }
 
-  async signMultiAgentTransaction(
-    transaction: TxnBuilderTypes.MultiAgentRawTransaction | TxnBuilderTypes.FeePayerRawTransaction
-  ): Promise<string | null> {
-    if (this._wallet && !("signMultiAgentTransaction" in this._wallet)) {
-      throw new WalletNotSupportedMethod(
-          `Multi agent & Fee payer transactions are not supported by ${this.wallet?.name}`
-      ).message;
-    }
-    try {
-      this.doesWalletExist();
-      const response = await (this._wallet as any).signMultiAgentTransaction(
-          transaction
-      );
-      return response;
-    } catch (error: any) {
-      const errMsg =
-          typeof error == "object" && "message" in error ? error.message : error;
-      throw new WalletSignTransactionError(errMsg).message;
-    }
-  }
-
   /**
   Event for when account has changed on the wallet
   @return the new account info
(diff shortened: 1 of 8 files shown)


0xmaayan (commented):
can you run changeset so we include it in next version release

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -501,4 +487,113 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
       throw new WalletSignMessageAndVerifyError(errMsg).message;
     }
   }
+
+  async prepRawTransaction(input: RawTransactionPrepPayload): Promise<AnyRawTransaction> {
```
0xmaayan, line 499: can we add some comments on what this function does

gregnazario: Oof, I forgot documentation in my mad rush, let me fix that

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -501,4 +487,113 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
+        gas_unit_price: input.options?.gas_unit_price?.toString(),
+        expiration_timestamp_secs: expiration_timestamp_secs.toString()
+      })
+    } else {
+      // TODO: Support multiagent without fee payer
+      throw new WalletNotSupportedMethod(`Multiagent without fee payer is not supported yet`).message;
+    }
+  }
+
+  async signRawTransaction(
```
0xmaayan, line 539: so this can be used for any transaction single/feepayer/agent?

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -501,4 +487,113 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
+    }
+    try {
+      this.doesWalletExist();
+      return await (this._wallet as any).signAndSubmitRawTransaction(input);
+    } catch (error: any) {
+      throw this.convertError(error);
+    }
+  }
+
+  private getClient(): Provider {
```
0xmaayan, line 578: what is the need for this function?

gregnazario: It's used above.

Basically, the sequence number and chain ID need to be pulled prior to generating the transaction.  This can alternatively be sent to the wallet instead of doing this.

packages/wallet-adapter-core/src/WalletCore.ts
gregnazario: Yes

0xmaayan (commented):
We should document those changes in https://github.com/aptos-labs/aptos-wallet-adapter/tree/main/packages/wallet-adapter-react

tbh, this becomes very complex, and with the new v2 support it would become even more complex. I think we need to start working on adapter v2 and then push wallets to support v2. We are going to end up with 3 different functions to submit a transaction

0xmaayan (approved):

gregnazario:
> We should document those changes in https://github.com/aptos-labs/aptos-wallet-adapter/tree/main/packages/wallet-adapter-react
> 
> tbh, this becomes very complex, and with the new v2 support it would become even more complex. I think we need to start working on adapter v2 and then push wallets to support v2. We are going to end up with 3 different functions to submit a transaction

Unfortunately there is always going to be minimally a 3 step process to submitting a transaction.

1. Generate the transaction (we can actually do this locally), but I've added it here for convenience so that users don't have to reimplement it.  The new SDK can just handle this on it's own.
2. Sign by one party (the primary signer, or the fee payer)
3. Sign and submit by the other party (primary signer or the fee payer)

This could be reduced down to mostly client logic, but this is to simplify the process since the fee payer key will have to be out of the client anyways.  If a wallet decides to sponsor the transactions itself, it can possibly skip over some of that to one call by the user.

0xmaayan:
> > We should document those changes in https://github.com/aptos-labs/aptos-wallet-adapter/tree/main/packages/wallet-adapter-react
> > tbh, this becomes very complex, and with the new v2 support it would become even more complex. I think we need to start working on adapter v2 and then push wallets to support v2. We are going to end up with 3 different functions to submit a transaction
> 
> Unfortunately there is always going to be minimally a 3 step process to submitting a transaction.
> 
> 1. Generate the transaction (we can actually do this locally), but I've added it here for convenience so that users don't have to reimplement it.  The new SDK can just handle this on it's own.
> 2. Sign by one party (the primary signer, or the fee payer)
> 3. Sign and submit by the other party (primary signer or the fee payer)
> 
> This could be reduced down to mostly client logic, but this is to simplify the process since the fee payer key will have to be out of the client anyways. If a wallet decides to sponsor the transactions itself, it can possibly skip over some of that to one call by the user.

I was mostly talking about `signAndSubmitTransaction`, `signAndSubmitRawTransaction`, `submitTransaction` (v2)

gregnazario:
> > > We should document those changes in https://github.com/aptos-labs/aptos-wallet-adapter/tree/main/packages/wallet-adapter-react
> > > tbh, this becomes very complex, and with the new v2 support it would become even more complex. I think we need to start working on adapter v2 and then push wallets to support v2. We are going to end up with 3 different functions to submit a transaction
> > 
> > 
> > Unfortunately there is always going to be minimally a 3 step process to submitting a transaction.
> > 
> > 1. Generate the transaction (we can actually do this locally), but I've added it here for convenience so that users don't have to reimplement it.  The new SDK can just handle this on it's own.
> > 2. Sign by one party (the primary signer, or the fee payer)
> > 3. Sign and submit by the other party (primary signer or the fee payer)
> > 
> > This could be reduced down to mostly client logic, but this is to simplify the process since the fee payer key will have to be out of the client anyways. If a wallet decides to sponsor the transactions itself, it can possibly skip over some of that to one call by the user.
> 
> I was mostly talking about `signAndSubmitTransaction`, `signAndSubmitRawTransaction`, `submitTransaction` (v2)

Yes, we need to collapse them into 1 for the new SDK and the future of the API.

We could just drop this and merge it in with the rest on the outside interface, but the wallets still need the inner interface.

gregnazario:
Closing in favor of SDK v2 bindings

PR #193, opened 2023-10-28
Supporting the new SDK with wallet adapter

xbtmatt:
### Description
This is the initial attempt at supporting the new SDK with the wallet adapter.

Namely, there is now a `submitTransaction` (TBD: new name) function where you can submit BCS inputs and it converts it to a BCS payload that's compatible with the v1 SDK. It then relays this call to the `signAndSubmitBCSTransaction` function that already exists. 

This facilitates signing and submitting a transaction using the v2 SDK without having to wait on wallets to implement any code for it.

The wallet adapter is still extremely limited in what it can do- this is an initial attempt at getting a single transaction to submit.

### Building
To use a local ts-sdk version with exported types (such as from the PR mentioned below), you must update the version in @aptos-labs/ts-sdk to link your local dependency, specifically in the

`apps/nextjs-example/package.json`
and
`packages/wallet-adapter-core/package.json`

```json
"@aptos-labs/ts-sdk": "^0.0.1",
"@aptos-labs/ts-sdk": "/YOUR/LOCAL/PATH/",
```

### Testing
go to the nextjs-example app, `yarn run dev`, and try to submit a transaction in the optional section at the bottom of the dapp. It will be for the v2 SDK

It should succeed. This relies on: [#139](https://github.com/aptos-labs/aptos-ts-sdk/pull/139) from the wallet adapter.


apps/nextjs-example/pages/index.tsx
```diff
@@ -7,8 +8,10 @@ import {
     Network,
     Provider,
     TxnBuilderTypes,
+    TypeTagParser,
```
0xmaayan: this is the old sdk, why we do new imports?

apps/nextjs-example/pages/index.tsx
```diff
@@ -467,6 +471,23 @@ function OptionalFunctionality() {
         );
     };
 
+    const onSubmitTransaction = async () => {
+        const response = await submitTransaction({
```
0xmaayan, line 477:
```suggestion
        if(!account){
          throw new Error(...)
        }
        const response = await submitTransaction({
        sender:account.address
```

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -371,23 +374,67 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
+      return response;
+    } catch (error: any) {
+      const errMsg =
+        typeof error == "object" && "message" in error ? error.message : error;
+      throw new WalletSignTransactionError(errMsg).message;
+    }
+  }
+
+  /**
+   * This function is for signing and submitting a transaction using the `@aptos-labs/ts-sdk` input types.
```
0xmaayan:
```suggestion
   * This function is for signing and submitting a transaction using the `@aptos-labs/ts-sdk` (aka SDK V2) input types.
```

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -371,23 +374,67 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
     }
   }
 
+  // TODO: Implement later. Wallets don't support this right now.
+  async signAnyTransaction(
```
0xmaayan: maybe should be out side of this PR scope?

apps/nextjs-example/pages/index.tsx
0xmaayan, line 477:
```suggestion
        const response: GenerateTransactionInput = await submitTransaction({
```

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -371,23 +374,67 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
+   * @param transactionInput the transaction input
+   * @param options max_gas_amount and gas_unit_limit
+   * @returns the response from the wallet's signAndSubmitBCSTransaction function
+   */
+  async submitTransaction(
+    transactionInput: GenerateTransactionInput,
+    options?: TransactionOptions,
+  ): Promise<any> {
+    const payloadData = transactionInput.data;
+    const newPayload = await TransactionBuilderTypes.generateTransactionPayload(payloadData);
```
0xmaayan: why do we need the `TransactionBuilderTypes`? can't we simply export `generateTransactionPayload` function?

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -371,23 +374,67 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
+    return response;
+  }
+
   async signMultiAgentTransaction(
     transaction: TxnBuilderTypes.MultiAgentRawTransaction | TxnBuilderTypes.FeePayerRawTransaction
   ): Promise<string | null> {
     if (this._wallet && !("signMultiAgentTransaction" in this._wallet)) {
       throw new WalletNotSupportedMethod(
-          `Multi agent & Fee payer transactions are not supported by ${this.wallet?.name}`
+        `Multi agent & Fee payer transactions are not supported by ${this.wallet?.name}`
```
0xmaayan:
```suggestion
        `Multi agent & Sponsor transactions are not supported by ${this.wallet?.name}`
```

packages/wallet-adapter-core/src/conversion.ts
```diff
@@ -0,0 +1,101 @@
+import { AccountAuthenticator, Deserializer, TransactionPayload, TransactionBuilderTypes, AnyRawTransaction, Network, TransactionPayloadScript, TransactionPayloadEntryFunction, TransactionPayloadMultisig } from "@aptos-labs/ts-sdk"
+import { BCS, HexString, TxnBuilderTypes, Types } from "aptos"
+import { NetworkInfo } from "./types";
+import { NetworkName } from "./constants";
+
+
+// we could serialize the entire thing and then deserialize
+// export const convertEntryFunction(entryFunction: EntryFunction): TxnBuilderTypes.EntryFunction {
```
0xmaayan: remove?

packages/wallet-adapter-core/src/conversion.ts
```diff
@@ -0,0 +1,101 @@
...
+        payloadType = "script_payload"
+    } else if (payload instanceof TransactionPayloadMultisig) {
+        payloadType = "multisig"
+        code = 
+    }
+}
+*/
+
+// new => old
+export function convertAuthenticator(authenticator: AccountAuthenticator): TxnBuilderTypes.AccountAuthenticator {
```
0xmaayan: looks like we dont really use this and some other functions in this file, do we need to keep them part of this PR?

apps/nextjs-example/pages/index.tsx
xbtmatt: Accident, removed!

apps/nextjs-example/pages/index.tsx
xbtmatt: That's not what it's returning- did you mean something else? Right now this returns the transaction response (but there's no guarantee because it's returning what the `signAndSubmitBCSTransaction` from the v1 adapter which is just `any`)

apps/nextjs-example/package.json
```diff
@@ -31,9 +32,9 @@
     "@welldone-studio/aptos-wallet-adapter": "^0.1.4",
     "antd": "^5.1.2",
     "aptos": "^1.19.0",
+    "@aptos-labs/ts-sdk": "^0.0.1",
```
0xmaayan: should be `^0.0.2`

apps/nextjs-example/pages/index.tsx
0xmaayan: yes my bad. this type should be what the function accepts. 
so something like
```
const input : GenerateTransactionInput = {
sender: account.address,
            data: {
                typeArguments: [new TypeTagParser(APTOS_COIN).parseTypeTag()],
                functionArguments: [AccountAddress.fromHexInput(account.address).toString(), 1]
}

const response = await submitTransaction({input})
```

packages/wallet-adapter-core/package.json
```diff
@@ -45,6 +45,7 @@
     "typescript": "^4.5.3"
   },
   "dependencies": {
+    "@aptos-labs/ts-sdk": "^0.0.1",
```
0xmaayan:
```suggestion
    "@aptos-labs/ts-sdk": "^0.0.2",
```

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -371,23 +374,46 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
+   * @param options max_gas_amount and gas_unit_limit
+   * @returns the response from the wallet's signAndSubmitBCSTransaction function
+   */
+  async submitTransaction(
+    transactionInput: InputGenerateTransactionData,
+    options?: TransactionOptions,
+  ): Promise<any> {
+    const payloadData = transactionInput.data;
+    const aptosConfig = new AptosConfig({network: convertNetwork(this._network)});
+    // TODO: Refactor this any, and remove the need for it by fixing the if ("bytecode" in data) stuff in `generateTransaction` in the v2 SDK
```
0xmaayan, line 392: can we add this task into our roadmap so we wont forget about it?

packages/wallet-adapter-react/src/useWallet.tsx
```diff
@@ -49,7 +50,7 @@ export interface WalletContextState {
   ): Promise<any>;
   signMessage(message: SignMessagePayload): Promise<SignMessageResponse | null>;
   signMessageAndVerify(message: SignMessagePayload): Promise<boolean>;
-
+  submitTransaction(transaction: InputGenerateTransactionData): Promise<any>;
```
gregnazario, line 53: Okay... we need to make this better.

`any` doesn't cut it anymore.  We should at least make it return a hash string or an object with a hash string so it can be extended accordingly.

gregnazario (commented):
Did we have anything else to get into the wallet adapter @0xmaayan for a release, otherwise, I suggest we get this in and just start iterating quickly on fixing what we want in the interface.

The sponsored transaction stuff as well as other pieces are quickly showing that they need a unified interface, and a split interface on the backend (as you mentioned in the other PR), and if we are going to deprecate the multiple functions, we'll need to drive towards that.

Thoughts?

0xmaayan:
> Did we have anything else to get into the wallet adapter @0xmaayan for a release, otherwise, I suggest we get this in and just start iterating quickly on fixing what we want in the interface.
> 
> The sponsored transaction stuff as well as other pieces are quickly showing that they need a unified interface, and a split interface on the backend (as you mentioned in the other PR), and if we are going to deprecate the multiple functions, we'll need to drive towards that.
> 
> Thoughts?

I think we need to move towards v2, it consolidates lots of the weird things we are doing here trying to support existing adapter, sdk v1 and sdk v2.
The new sdk provide fewer and easier interfaces to generate/sign/submit transaction

gregnazario:
> > Did we have anything else to get into the wallet adapter @0xmaayan for a release, otherwise, I suggest we get this in and just start iterating quickly on fixing what we want in the interface.
> > The sponsored transaction stuff as well as other pieces are quickly showing that they need a unified interface, and a split interface on the backend (as you mentioned in the other PR), and if we are going to deprecate the multiple functions, we'll need to drive towards that.
> > Thoughts?
> 
> I think we need to move towards v2, it consolidates lots of the weird things we are doing here trying to support existing adapter, sdk v1 and sdk v2. The new sdk provide fewer and easier interfaces to generate/sign/submit transaction

Let's get this in @xbtmatt, and let's get it rolling on wallet adapter V2 with more functionality.

packages/wallet-adapter-core/src/WalletCore.ts
xbtmatt: added it!

gregnazario (approved):
Please make an interface for the output of `submitTransaction` of `{hash: string}`

0xmaayan (approved):

0xmaayan:
@xbtmatt please run `pnpm run changeset` so we can add it to version packages PR

PR #195, opened 2023-10-30
Moved `aptos`, `react` and `react-dom` to `peerDependencies`

hardsetting:
When a shared dependency like `aptos` is defined under `dependencies`, the package manager can decide to pull different bundles for each package that depends on it.

Defining it under `peerDependencies` is telling the package manager "I need this package in any of these version, but I'll let the root package.json decide the version", ensuring (when possible) a single bundle under `node_modules`.

This is particularly important for dependencies that share classes and instances e.g. `axios`, `aptos`, `react` ...
If there's multiple bundles (even if the versions are compatible or the same), operators like `instanceof` are not going to work as expected

### DIFF
--- a/.changeset/three-moose-eat.md
+++ b/.changeset/three-moose-eat.md
@@ -1,0 +1,8 @@
+---
+"@aptos-labs/wallet-adapter-ant-design": patch
+"@aptos-labs/wallet-adapter-core": patch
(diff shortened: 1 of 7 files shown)


gregnazario (approved):

0xmaayan:
Can you run changeset ? so we can release a new version with it

AaricChen (approved):

PR #197, opened 2023-11-10
Wallet adapter v2

0xmaayan:
Wallet Adapter V2

- Dapp to support SDK V2 and use only @aptos-labs/ts-sdk package. 
- Add 3 different transaction flows - Single signer, Fee payer, Multi Agent. See attached video. 
   - Single signer has the same functionality with the option to `signAndSubmitTransction`, `signMessage`, `signMessageAndVerify`. For `signTransaction` we had to keep both V1 and V2 "ways" as we can't really change the function signature without introducing wallet plugin breaking changes. We can technically just use a new function with a different name only for V2 but rather to use the same function to support both v1 and v2 to easy ecosystem adoption 
   - Fee payer flow supports `sign as sender`, then need to switch wallet to `sign as sponsor`, and then submit transaction (everything implemented in the V2 way)
   - Multi agent flow supports `sign as sender`, then you would get a private key you can import to the wallet to `sign as secondary signer`, and then submit transaction (everything implemented in the V2 way)

Note: I chose this way so users can experience the use of signing with a wallet and can experience the e2e flow with a wallet.

 - Wallet core conversion layer from `signAndSubmitTransaction` with V2 inputs -> V1 `signAndSubmitTransaction` or V2 `signAndSubmitBCSTransaction`
 - Wallet core conversion layer to return V2 `AccountAuthenticator` from `signTransaction`
 - `submitTransaction` to only support V2 expected flow

https://github.com/aptos-labs/aptos-wallet-adapter/assets/29798064/580d2ad4-a514-479c-8de5-2c12a66d2e11

### DIFF
--- a/apps/nextjs-example/components/transactionFlow/MultiAgent.tsx
+++ b/apps/nextjs-example/components/transactionFlow/MultiAgent.tsx
@@ -46,7 +46,7 @@ export default function MultiAgentTransaction({
     const generate = async (): Promise<AnyRawTransaction> => {
       const transactionToSign = await aptosClient(
         network?.name.toLowerCase()
-      ).generateTransaction({
+      ).build.multiAgentTransaction({
         sender: account.address,
         secondarySignerAddresses: [secondarySigner.accountAddress],
         data: {
--- a/apps/nextjs-example/components/transactionFlow/MultiAgent.tsx
+++ b/apps/nextjs-example/components/transactionFlow/MultiAgent.tsx
@@ -1,0 +1,179 @@
+import {
+  Account,
+  AccountAuthenticator,
+  AnyRawTransaction,
+} from "@aptos-labs/ts-sdk";
+import { useWallet } from "@aptos-labs/wallet-adapter-react";
+
+import { useState, useEffect } from "react";
+import { aptosClient } from "../../utils";
+import { useAlert } from "../AlertProvider";
+import Button from "../Button";
+import Col from "../Col";
+import Row from "../Row";
+export const APTOS_COIN = "0x1::aptos_coin::AptosCoin";
+type MultiAgentTransactionProps = {
+  isSendableNetwork: (connected: boolean, network?: string) => boolean;
+};
+
+export default function MultiAgentTransaction({
+  isSendableNetwork,
+}: MultiAgentTransactionProps) {
+  const { connected, account, network, signTransaction, submitTransaction } =
+    useWallet();
+
+  const [secondarySignerAccount, setSecondarySignerAccount] =
+    useState<Account>();
+  const [transactionToSubmit, setTransactionToSubmit] =
+    useState<AnyRawTransaction | null>(null);
+  const { setSuccessAlertHash } = useAlert();
+
+  const [senderAuthenticator, setSenderAuthenticator] =
+    useState<AccountAuthenticator>();
+  const [secondarySignerAuthenticator, setSecondarySignerAuthenticator] =
+    useState<AccountAuthenticator>();
+
+  let sendable = isSendableNetwork(connected, network?.name);
+
+  useEffect(() => {
+    if (!sendable) return;
+    if (!account) return;
+    if (!network) return;
+
+    const secondarySigner = Account.generate();
+
+    // Generate a raw transaction using the SDK
+    const generate = async (): Promise<AnyRawTransaction> => {
+      const transactionToSign = await aptosClient(
+        network?.name.toLowerCase()
+      ).generateTransaction({
+        sender: account.address,
+        secondarySignerAddresses: [secondarySigner.accountAddress],
+        data: {
+          function: "0x1::coin::transfer",
+          typeArguments: [APTOS_COIN],
+          functionArguments: [account.address, 1], // 1 is in Octas
+        },
+      });
+      return transactionToSign;
+    };
+
+    // Fund secondary signer account to create it on chain
+    const fundSecondarySigner = async () => {
+      await aptosClient(network.name.toLowerCase()).fundAccount({
+        accountAddress: secondarySigner.accountAddress.toString(),
+        amount: 100_000_000,
+      });
+    };
+
+    // Fund current connected account to create it on chain
+    const fundCurrentAccount = async () => {
+      await aptosClient(network.name.toLowerCase()).fundAccount({
+        accountAddress: account.address,
+        amount: 100_000_000,
+      });
+    };
+
+    // We fund the secondary signer account to create it on chain
+    // We fund the current connected account to create it on chain
+    // Then we generate the transaction with the current connected account
+    // as the sender and the secondary signer account
+    fundSecondarySigner().then(() => {
+      setSecondarySignerAccount(secondarySigner);
+      fundCurrentAccount().then(() => {
+        generate().then((transactionToSign) =>
+          setTransactionToSubmit(transactionToSign)
+        );
+      });
+    });
+  }, [network, account, sendable]);
+
+  const onSenderSignTransaction = async () => {
+    if (!transactionToSubmit) {
+      throw new Error("No Transaction to sign");
+    }
+    try {
+      const authenticator = await signTransaction(transactionToSubmit);
+      setSenderAuthenticator(authenticator);
+    } catch (error) {
+      console.error(error);
+    }
+  };
+
+  const onSecondarySignerSignTransaction = async () => {
+    if (!transactionToSubmit) {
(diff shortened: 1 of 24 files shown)


packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -266,86 +281,141 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
+
+      // Convert retuned bcs serialized SignedTransaction into V2 AccountAuthenticator
+      const deserializer1 = new BCS.Deserializer(response);
+      const deserializedSignature =
+        TxnBuilderTypes.SignedTransaction.deserialize(deserializer1);
+      const transactionAuthenticator =
+        deserializedSignature.authenticator as TxnBuilderTypes.TransactionAuthenticatorEd25519;
+
+      const publicKey = transactionAuthenticator.public_key.value;
+      const signature = transactionAuthenticator.signature.value;
```
0xmaayan, line 429: this is where it is a bit weird. When trying to use this publicKey, signature and message to verifyMessage - verification fails. Failure is happening in both SDK V1 and V2.
```
const ed25519 = new Ed25519PublicKey(publicKey);
const verify = ed25519.verifyMessage(response, signature);
```

Consider the `signTransaction` function is being used with wallets we can assume it works? so 2 failure options
1. verifyMessage is wrong
2. something in the deserialization process is wrong

cc @xbtmatt @hardsetting

packages/wallet-adapter-react/src/useWallet.tsx
```diff
@@ -36,32 +39,27 @@ export interface WalletContextState {
   disconnect(): void;
   wallet: WalletInfo | null;
   wallets: ReadonlyArray<Wallet>;
-  signAndSubmitTransaction<T extends Types.TransactionPayload>(
-    transaction: T,
-    options?: TransactionOptions
-  ): Promise<any>;
-  signAndSubmitBCSTransaction<T extends TxnBuilderTypes.TransactionPayload>(
-    transaction: T,
-    options?: TransactionOptions
```
gregnazario, line 45: Okay, so this will no longer work.  Good to know for future dapps to be put in a changelog

packages/wallet-adapter-react/src/WalletProvider.tsx
```diff
@@ -74,90 +78,68 @@ export const AptosWalletAdapterProvider: FC<AptosWalletProviderProps> = ({
...
   ) => {
     try {
-      return await walletCore.submitTransaction(transaction);
+      return await walletCore.signAndSubmitTransaction(transaction, options);
     } catch (error: any) {
       if (onError) onError(error);
-      else throw error;
-      return false;
+      return Promise.reject(error);
     }
```
gregnazario, line 142: Awesome!  finally no weird false

packages/wallet-adapter-core/src/WalletCoreV1.ts
```diff
@@ -0,0 +1,86 @@
...
+    transaction: Types.TransactionPayload,
+    wallet: Wallet | null,
+    options?: TransactionOptions
+  ): Promise<Uint8Array | null> {
+    try {
+      const response = await (wallet as any).signTransaction(
+        transaction,
+        options
+      );
+      console.log("response v1", response);
```
gregnazario: Should remove the console logs?  Unless it always did that

packages/wallet-adapter-core/src/WalletCoreV1.ts
```diff
@@ -0,0 +1,86 @@
...
+      return response;
+    } catch (error: any) {
+      const errMsg =
+        typeof error == "object" && "message" in error ? error.message : error;
```
gregnazario, line 81: Any reason this isn't `typeof error === "object"`.  We should put this in a utility function, since it's common among all of these for error handling.

packages/wallet-adapter-core/src/WalletCoreV1.ts
```diff
@@ -0,0 +1,86 @@
...
+   @param transaction
+   @param options max_gas_amount and gas_unit_limit
+   @return response from the wallet's signTransaction function
+   @throws WalletSignTransactionError
+   */
+  async signTransaction(
+    transaction: Types.TransactionPayload,
+    wallet: Wallet | null,
+    options?: TransactionOptions
+  ): Promise<Uint8Array | null> {
```
gregnazario, line 72: sigh... someday we'll get rid of the nulls.  It's hard since it requires to handle the arbitrary calls.

packages/wallet-adapter-core/src/WalletCoreV1.ts
```diff
@@ -0,0 +1,86 @@
...
+  /**
+  Sign and submit an entry (not bcs serialized) transaction type to chain.
+  @param transaction a non-bcs serialized transaction
+  @param options max_gas_amount and gas_unit_limit
+  @return response from the wallet's signAndSubmitTransaction function
+  @throws WalletSignAndSubmitMessageError
+  */
+  async signAndSubmitTransaction(
+    transaction: Types.TransactionPayload,
+    wallet: Wallet | null,
```
gregnazario: Is there a way to ensure the Wallet is not null?

apps/nextjs-example/components/transactionFlow/MultiAgent.tsx
```diff
@@ -0,0 +1,179 @@
...
+            <h3>Secondary Signer details</h3>
+          </Col>
+          <Col>
+            <p>Private Key: {secondarySignerAccount.privateKey.toString()}</p>
+            <p>Public Key: {secondarySignerAccount.publicKey.toString()}</p>
+            <p>Address: {secondarySignerAccount.accountAddress.toString()}</p>
+          </Col>
+        </Row>
+      )}
+    </>
```
gregnazario, line 177: Awesome

apps/nextjs-example/components/transactionFlow/Sponsor.tsx
```diff
@@ -0,0 +1,134 @@
...
+    }
+    try {
+      const authenticator = await signTransaction(transactionToSubmit);
+      if (asSponsor) {
+        setFeepayerAuthenticator(authenticator);
+      } else {
+        setSenderAuthenticator(authenticator);
+      }
+    } catch (error) {
+      console.error(error);
```
gregnazario: We may want to approach the fee payer as a different account to show the example, similar to multi-agent.  But, this works for now.

apps/nextjs-example/components/transactionFlow/Sponsor.tsx
```diff
@@ -0,0 +1,134 @@
...
+    if (!transactionToSubmit) {
+      throw new Error("No Transaction to sign");
+    }
+    try {
+      const authenticator = await signTransaction(transactionToSubmit);
+      if (asSponsor) {
+        setFeepayerAuthenticator(authenticator);
+      } else {
+        setSenderAuthenticator(authenticator);
+      }
```
gregnazario: How does it sign properly for fee payer?  Signing would have to happen after the transaction has been changed to include the fee payer.

The fee payer has to sign with their address in the fee payer slot.

apps/nextjs-example/package.json
```diff
@@ -11,6 +11,7 @@
     "lint": "next lint"
   },
   "dependencies": {
+    "@aptos-labs/ts-sdk": "^0.0.5",
```
gregnazario: Bump it to 0.0.6!

packages/wallet-adapter-core/package.json
```diff
@@ -45,12 +45,12 @@
...
   "dependencies": {
-    "@aptos-labs/ts-sdk": "^0.0.3",
     "buffer": "^6.0.3",
     "eventemitter3": "^4.0.7",
     "tweetnacl": "^1.0.3"
   },
   "peerDependencies": {
-    "aptos": "^1.19.0"
+    "aptos": "^1.19.0",
+    "@aptos-labs/ts-sdk": "^0.0.5"
```
gregnazario: similarly

packages/wallet-adapter-core/src/WalletCore.ts
gregnazario: We can look together at the derserialized bytes to see what's going on here, but it should be pretty simple to separate and determine.

The other possibility is if this is a signed message and not a transaction, is it might not have all of the data for the message to verify.

gregnazario (approved):

PR #221, opened 2023-12-18
Return false from `areBCSArguments` function if array is empty

0xmaayan:
`areBCSArguments` function will return `true` if `args` is empty (`every` returns true for an empty array). That causes the adapter to use `signAndSubmitBCSTransaction` when `functionArguments` is empty. We want to avoid it, and use `signAndSubmitTransaction` since all wallets support this function and not the former one.

### DIFF
--- a/packages/wallet-adapter-core/src/utils/helpers.ts
+++ b/packages/wallet-adapter-core/src/utils/helpers.ts
@@ -38,5 +38,8 @@ export function generalizedErrorMessage(error: any): string {
 // Serializable, so if each argument is of an instance of a class
 // the extends Serializable - we know these are BCS arguments
 export const areBCSArguments = (args: any): boolean => {
+  // `every` returns true if the array is empty, so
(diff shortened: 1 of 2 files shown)


packages/wallet-adapter-core/src/utils/helpers.ts
```diff
@@ -38,5 +38,8 @@ export function generalizedErrorMessage(error: any): string {
 // Serializable, so if each argument is of an instance of a class
 // the extends Serializable - we know these are BCS arguments
 export const areBCSArguments = (args: any): boolean => {
+  // `every` returns true if the array is empty, so
+  // first check the array length
+  if (args.length === 0) return false;
   return args.every((arg: any) => arg instanceof Serializable);
 };
```
gregnazario: Is args an array of any?  We should probably fix the input argument to the function.

gregnazario (approved):

PR #224, opened 2024-01-04
Update to latest sdk version `1.3.0`

0xmaayan:
This PR updates ts sdk to the latest version where we use `number` types for transaction options parameters. https://github.com/aptos-labs/aptos-ts-sdk/pull/248

With this change the legacy code still converts transaction option parameters to `BigInt` for wallet backward compatibility, but for any usage with the new sdk we use the `number` type.

### DIFF
--- a/.changeset/dirty-guests-bathe.md
+++ b/.changeset/dirty-guests-bathe.md
@@ -1,0 +1,6 @@
+---
+"@aptos-labs/wallet-adapter-core": minor
+"@aptos-labs/wallet-adapter-nextjs-example": minor
+---
+
+Update @aptos-labs/ts-sdk package version to 1.3.0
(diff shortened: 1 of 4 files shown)


gregnazario (dismissed):

gregnazario (commented):
What about the version file?

0xmaayan:
> What about the version file?

what version file?

gregnazario:
> > What about the version file?
> 
> what version file?

Oh woops, I thought this was the SDK.

We need to also update SDKv1 into here to fix it entirely

gregnazario (approved):

0xmaayan:
> > > What about the version file?
> > 
> > 
> > what version file?
> 
> Oh woops, I thought this was the SDK.
> 
> We need to also update SDKv1 into here to fix it entirely

I dont think so. This is for a project that uses the new sdk, legacy code functions the same

PR #243, opened 2024-03-18
feat: Change the return value of the connect function to Promise<AccountInfo | null>

livetOwO:
A Pull Request has been created to enable various operations without the necessity of using useEffect after invoking the connect function in the wallet-adapter-react library.

related docs: 
- https://react.dev/reference/rules/components-and-hooks-must-be-pure#side-effects-must-run-outside-of-render
- https://aptos.dev/standards/wallets/#connection-apis

### DIFF
--- a/.changeset/real-moose-own.md
+++ b/.changeset/real-moose-own.md
@@ -1,0 +1,5 @@
+---
+"@aptos-labs/wallet-adapter-react": minor
+---
+
+Change the return value of the connect function to Promise<AccountInfo | null>
--- a/packages/wallet-adapter-react/src/WalletProvider.tsx
+++ b/packages/wallet-adapter-react/src/WalletProvider.tsx
@@ -66,6 +66,7 @@ export const AptosWalletAdapterProvider: FC<AptosWalletProviderProps> = ({
     try {
       setIsLoading(true);
       await walletCore.connect(walletName);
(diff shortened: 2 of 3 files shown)


0xmaayan:
Thank you for the contribution. 

> A Pull Request has been created to enable various operations without the necessity of using useEffect after invoking the connect function in the wallet-adapter-react library.

The React provider package [updates](https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-react/src/WalletProvider.tsx#L167) the `account` state on a `connect` event and that bubbles down to the end user.
The end user (dapp) should use the `account` state provided by the provider to manage the current connected account

> * https://aptos.dev/standards/wallets/#connection-apis

This is a Wallet standard, i.e a `Wallet` must return `AccountInfo` on `connect` method

Therefore, I am not completely convinced this change is needed. Could you clarify more the use case and the need for this change?

livetOwO:
@0xmaayan 
Thank you for the comment

My service collects from amplitude and Google Analytics that the user connected their wallet immediately after connecting.
At the same time, it requests a session to the service using signMessage.
Because of this, I need to return a promise or have a callback function.
The reason I need to return the account is because I don't want to use signMessage without any usage effect.

PR #244, opened 2024-03-21
Add AIP-62 standard support

0xmaayan:
This PR adds support for the new wallet standard (AIP-62) while keeping support on the existing wallet standard.

**Live demo dapp with this PR logic** 
https://aptos-labs.github.io/aptos-wallet-adapter/nextjs-example-testing/
Notes:
- Nightly is an AIP-62 standard compatible wallet

**TL;DR**
- Change adapter expected wallet plugins array to optional
- Add logic to support new wallet standard following AIP-62
- Create and expose all existing and new standard wallets to the dapp user
- Dapps that use the `wallet-adapter-react` package should simply upgrade the version
- Dapps that use the `wallet-adapter-antd-design` or `wallet-adapter-mui-design` packages should simply upgrade the version 
- Dapps that use the native `wallet-adapter-core` package would need to upgrade the version and potentially some small modifications 
- Dapps can use both standard support versions, or only the new standard version (by not providing any wallet plugins)
- Wallets that use the `wallet-adapter-core` package would need to upgrade the version

**Existing standard (adapter wallet plugin)**
Currently, a dapp needs to install adapter wallet plugin dependency and to provide it to the adapter. The wallet adapter keeps this options BUT changes the expected provided plugins array to be optional.

**New standard wallet**
The new wallet standard operates by dapp<>wallet event communication. Therefore, the dapp itself doesnt need to install or maintain any wallet dependency. The wallet adapter adds logic to support the new standard while maintaining support with the existing standard. The wallet adapter core handles all existing and new standard wallets detections and put them all in an array that eventually will be visible to the dapp user.

Following this implementation, dapps and wallets that follow the current implementation recommendation should not do any major changes but simply upgrade the adapter version they use.


packages/wallet-adapter-react/src/WalletProvider.tsx
```diff
@@ -210,11 +209,28 @@ export const AptosWalletAdapterProvider: FC<AptosWalletProviderProps> = ({
...
+        // Return a new object with updated value
+        return { ...wallet, readyState: updatedWallet.readyState };
+      }
+      return wallet;
+    });
+    setWallets(updatedWallets);
+  };
+
+  const handleStandardWalletsAdded = (standardWallet: Wallet) => {
+    // Manage current wallet state by removing optonal duplications
```
gregnazario:
```suggestion
    // Manage current wallet state by removing optional duplications
```

packages/wallet-adapter-core/src/ans.ts
```diff
@@ -1,13 +1,16 @@
...
 const ChainIdToAnsContractAddressMap: Record<string, string> = {
   "1": "mainnet", // mainnet
   "2": "testnet", // testnet
 };
 
 export const getNameByAddress = async (
-  chainId: string,
-  address: string
+  chainId: string | number,
+  address: string | AccountAddress
```
gregnazario: Nice, finally.

We might want to actually add get primary name by address as well as the main replacement for this (or use the V2 SDK directly)

apps/nextjs-example/utils/standardWallet.ts
```diff
@@ -0,0 +1,238 @@
...
+  AptosOnNetworkChangeMethod,
+  AptosFeatures,
+  UserResponseStatus,
+} from "@aptos-labs/wallet-standard";
+
+/**
+ * A class to create a mock wallet for demonstration a wallet
+ * implementation compatible with Aptos AIP-62 Wallet Standard
+ */
+export class MyWalletAccount implements AptosWalletAccount {
```
gregnazario, line 38: Love the test wallet

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -98,6 +139,117 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
     });
   }
 
+  private fetchAptosWallets() {
+    let { aptosWallets, on } = getAptosWallets();
+    this.setWallets(aptosWallets);
+
+    if (typeof window === "undefined") return;
```
gregnazario, line 145: Just thoughts for later, we should probably put these into a reusable function, since we check that the window is undefined lots of places.

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -129,28 +293,123 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
+        const connectStandardAccount =
+          account as UserResponse<StandardAccountInfo>;
+        if (connectStandardAccount.status === UserResponseStatus.REJECTED) {
+          this._connecting = false;
+          throw new WalletConnectionError("User has rejected the request")
+            .message;
+        }
+        // account is of type
+        this._account = {
+          address: connectStandardAccount.args.address.toString(),
```
gregnazario, line 348: Is there a reason not to have it be `AccountAddress` or is that just for backwards compatibility?

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -213,16 +475,22 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
 
-    // check if we are in a redirectable view (i.e on mobile AND not in an in-app browser) and
-    // since wallet readyState can be NotDetected, we check it before the next check
+    // Check if we are in a redirectable view (i.e on mobile AND not in an in-app browser)
     if (isRedirectable()) {
       // use wallet deep link
+      if (selectedWallet.isAIP62Standard && selectedWallet.openInMobileApp) {
+        selectedWallet.openInMobileApp();
+        return;
+      }
```
gregnazario, line 499: Awesome mobile redirection

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -296,65 +564,57 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
-      const { hash, ...output } = response;
-      return { hash, output };
+      }
+
+      // If wallet does not support signAndSubmitTransaction
+      // the adapter will sign and submit it for the dapp.
+      // Note: This should happen only for AIP-62 standard compatible wallets since
+      // signAndSubmitTransaction is not a required function implementation
+      const transaction = await aptos.transaction.build.simple({
+        sender: this._account.address.toString(),
```
gregnazario: Address is already a string right?  probably can skip toString()

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -377,110 +637,146 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
+            return await this.walletStandardCore.signTransaction(
+              transactionOrPayload,
+              this._wallet,
+              asFeePayer
+            );
+          } else {
+            // else it means dapp passes legacy sdk v1 input data
+            // and the current connected wallet does not support it
+            throw new WalletNotSupportedMethod(
+              `Wrong transaction input data was porivded, ${this.wallet?.name} expects AnyRawTransaction input type.
```
gregnazario:
```suggestion
              `Invalid transaction input data was provided, ${this.wallet?.name} expects AnyRawTransaction type. 
```

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -377,110 +637,146 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
+              transactionOrPayload,
+              this._wallet,
+              asFeePayer
+            );
+          } else {
+            // else it means dapp passes legacy sdk v1 input data
+            // and the current connected wallet does not support it
+            throw new WalletNotSupportedMethod(
+              `Wrong transaction input data was porivded, ${this.wallet?.name} expects AnyRawTransaction input type. 
+              Please upgrade to the new Aptos TS SDK`
```
gregnazario: A link might be more helpful as we push for more aggressive deprecation.

```suggestion
              Please upgrade to the latest Aptos TS SDK https://github.com/aptos-labs/aptos-ts-sdk`
```

packages/wallet-adapter-core/src/LegacyWalletPlugins/types.ts
```diff
@@ -96,32 +108,48 @@ export interface AdapterPluginProps<Name extends string = string> {
...
+  // Comaptible with AIP-62 standard wallet
+  openInMobileApp?: () => void;
   connect(): Promise<any>;
   disconnect: () => Promise<any>;
   network: () => Promise<any>;
-  signAndSubmitTransaction(
-    transaction: Types.TransactionPayload | InputGenerateTransactionData,
+  signAndSubmitTransaction?(
+    transaction:
+      | Types.TransactionPayload
```
gregnazario, line 119: Kinda weird, didn't know you could put a `|` before and have nothing before it

gregnazario (commented):
It generally looks good, I would suggest we get a wallet person to look over it as well

packages/wallet-adapter-core/src/LegacyWalletPlugins/types.ts
0xmaayan: it is just a format thing, we are having it across other ts tools

packages/wallet-adapter-core/src/WalletCore.ts
0xmaayan: The wallet holds and returns it as `AccountAddress` (i.e `connectStandardAccount` is of type `AccountAddress`), as for the adapter we
1) want to keep backwards compatibility and 2) dont think it makes sense to returns to the dapp a `AccountAddress` type and have it resolves everything, the adapter should do it for them

apps/nextjs-example/components/WalletButtons.tsx
```diff
@@ -6,22 +6,26 @@ import {
...
   );
 };
 
-const WalletView = (wallet: Wallet) => {
-  const { connect } = useWallet();
-  const { setErrorAlertMessage } = useAlert();
+const WalletView = (
+  wallet: Wallet,
+  connect: (walletName: WalletName) => void,
+  setErrorAlertMessage: Dispatch<SetStateAction<ReactNode>>
```
hardsetting, line 27: Curious, why was this change needed?
To re-use the alert?

packages/wallet-adapter-core/package.json
```diff
@@ -44,13 +44,14 @@
     "tsup": "^5.10.1",
     "typescript": "^4.5.3"
   },
+  "peerDependencies": {
+    "@aptos-labs/ts-sdk": "^1.10.0"
+  },
```
hardsetting: was moving peer deps to the top intentional?

packages/wallet-adapter-core/src/constants.ts
```diff
@@ -23,3 +23,8 @@ export enum NetworkName {
   Testnet = "testnet",
   Devnet = "devnet",
 }
+
+export const ChainIdToAnsContractAddressMap: Record<string, string> = {
+  "1": "mainnet", // mainnet
+  "2": "testnet", // testnet
+};
```
hardsetting: it says "chain id to contract address" but the map points to a network name?

packages/wallet-adapter-core/src/AIP62StandardWallets/WalletStandard.ts
```diff
@@ -0,0 +1,176 @@
...
+   * Signs and submits a transaction to chain
+   *
+   * @param transactionInput InputTransactionData
+   * @returns PendingTransactionResponse
+   */
+  async signAndSubmitTransaction(
+    transactionInput: InputTransactionData,
+    aptos: Aptos,
+    account: AccountInfo,
+    wallet: Wallet
```
hardsetting, line 48: we're still passing a "legacy" wallet to the new WalletStandardCore?

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -98,6 +138,117 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
   }
 
+  private fetchAptosWallets() {
+    let { aptosWallets, on } = getAptosWallets();
+    this.setWallets(aptosWallets);
+
+    if (typeof window === "undefined") return;
+    // Adds an event listener for new wallets that get registered after the dapp has been loaded,
+    // receiving an unsubscribe function, which it can later use to remove the listener
+    const that = this;
```
hardsetting, line 148: 😆

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -98,6 +138,117 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
+    const removeRegisterListener = on("register", function () {
+      let { aptosWallets } = getAptosWallets();
+      that.setWallets(aptosWallets);
+    });
+
+    const removeUnregisterListener = on("unregister", function () {
+      let { aptosWallets } = getAptosWallets();
+      that.setWallets(aptosWallets);
+    });
```
hardsetting, line 157: Maybe there's no need to worry about it at this point, but we need to at least give users a way to unregister the listeners.

What happens if you instantiate a WalletCore, then destroy it and instantiate a new one?
Not sure when that would happen, but it could.
Like when using react's `StrictMode`

hardsetting: No action item for this PR necessarily, but wanted to surface this jic

packages/wallet-adapter-react/src/WalletProvider.tsx
```diff
@@ -57,7 +57,7 @@ export const AptosWalletAdapterProvider: FC<AptosWalletProviderProps> = ({
   // https://github.com/aptos-labs/aptos-wallet-adapter/issues/94
   const [isLoading, setIsLoading] = useState<boolean>(true);
 
-  const walletCore = useMemo(() => new WalletCore(plugins), []);
+  const walletCore = useMemo(() => new WalletCore(plugins ?? []), []);
```
hardsetting, line 60: Shouldn't this have `plugins` in the dependency array?
Or is that intended

packages/wallet-adapter-react/src/WalletProvider.tsx
```diff
@@ -223,14 +239,16 @@ export const AptosWalletAdapterProvider: FC<AptosWalletProviderProps> = ({
...
     return () => {
       walletCore.off("connect", handleConnect);
       walletCore.off("disconnect", handleDisconnect);
       walletCore.off("accountChange", handleAccountChange);
       walletCore.off("networkChange", handleNetworkChange);
       walletCore.off("readyStateChange", handleReadyStateChange);
+      walletCore.off("standardWalletsAdded", handleStandardWalletsAdded);
     };
-  }, [...wallets, connected]);
+  }, [wallets, connected]);
```
hardsetting, line 251: oh that was scary good catch

hardsetting (approved):
Just a few questions to make sure I understand correctly, but otherwise looks good!

apps/nextjs-example/components/WalletButtons.tsx
0xmaayan: React complains on re-rendering thing, to solve it was to move all hooks to the top of the component (can't remember the exact error react gave)

packages/wallet-adapter-core/package.json
0xmaayan: no... just formatting I guess..

packages/wallet-adapter-core/src/AIP62StandardWallets/WalletStandard.ts
0xmaayan: it is not a "legacy" wallet but a wallet adapter Wallet type, this is to not introduce any breaking changes for dapps

packages/wallet-adapter-core/src/constants.ts
0xmaayan: yep - will update

packages/wallet-adapter-react/src/WalletProvider.tsx
0xmaayan: this is to support only the new standard - meaning if a dapp wants to use then ew standard (and not the legacy one) they should not pass in any wallet plugin

packages/wallet-adapter-core/src/AIP62StandardWallets/WalletStandard.ts
0xmaayan: we could do it differently and
- convert `wallet` to `AptosStandardWallet`
- Pass it to the new standard `signAndSubmitTransaction` function
- and then do `wallet.features["aptos:signAndSubmitTransaction"]?.signAndSubmitTransaction(
        transaction
      )`

but we already converted each `AptosStandardWallet` wallet into legacy `Wallet` type so we kind of convert it now back.

I think once we deprecate any legacy wallet plugin, we will have to change adapter types and therefore input and output types - that would introduce a breaking change

packages/wallet-adapter-react/src/WalletProvider.tsx
0xmaayan: I just re-read you comment. I don think having it in the dependency array is needed as this array is being passed one time when using the adapter react provider.

PR #246, opened 2024-03-27
Remove Nightly package dependency

0xmaayan:
Now that we have migrated to the new wallet adapter, a dapp doesnt need to hold a package depndency of a wallet that is compatible with the new standard

### DIFF
--- a/apps/nextjs-example/components/AppContext.tsx
+++ b/apps/nextjs-example/components/AppContext.tsx
@@ -4,7 +4,6 @@ import { FaceWallet } from "@haechi-labs/face-aptos-adapter-plugin";
 import { FewchaWallet } from "fewcha-plugin-wallet-adapter";
 import { FlipperWallet } from "@flipperplatform/wallet-adapter-plugin";
 import { MartianWallet } from "@martianwallet/aptos-wallet-adapter";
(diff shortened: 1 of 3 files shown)


0x-j (approved):

0x-j:
does this mean dapp integrating wallet adaptor no longer need to add petra, okx, martian, etc. they only need to add wallet adaptor as dependency?

0xmaayan:
> does this mean dapp integrating wallet adaptor no longer need to add petra, okx, martian, etc. they only need to add wallet adaptor as dependency?

correct. as long as the wallet itself has migrated to the new standard

0x-j:
> > does this mean dapp integrating wallet adaptor no longer need to add petra, okx, martian, etc. they only need to add wallet adaptor as dependency?
> 
> correct. as long as the wallet itself has migrated to the new standard

wow nice, finally i dont need to have 10 wallets in dependency

0xmaayan:
> > > does this mean dapp integrating wallet adaptor no longer need to add petra, okx, martian, etc. they only need to add wallet adaptor as dependency?
> > 
> > 
> > correct. as long as the wallet itself has migrated to the new standard
> 
> wow nice, finally i dont need to have 10 wallets in dependency

YES!

PR #250, opened 2024-04-05
Implement GA4

0xmaayan:
Adds support for Google Analytics 4 to track event in the wallet adapter.

### DIFF
--- a/.changeset/tasty-humans-love.md
+++ b/.changeset/tasty-humans-love.md
@@ -1,0 +1,7 @@
+---
+"@aptos-labs/wallet-adapter-ant-design": minor
+"@aptos-labs/wallet-adapter-mui-design": minor
+"@aptos-labs/wallet-adapter-core": minor
+---
+
+Implement GA4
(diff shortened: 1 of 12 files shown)


BriungRi (commented):
Do these event recordings need to be wrapped in `try`/`catch`? I'm wondering if some of these requests will block or fail if the user has AdBlock enabled.

0xmaayan:
> Do these event recordings need to be wrapped in `try`/`catch`? I'm wondering if some of these requests will block or fail if the user has AdBlock enabled.

interesting... not sure. I could probably put `recordEvent` in a try/catch block

0xmaayan:
> Do these event recordings need to be wrapped in `try`/`catch`? I'm wondering if some of these requests will block or fail if the user has AdBlock enabled.

ok I tested it with an ad blocker (using Brave browser) and I dont think we need to use try/catch block as it doesnt break the app if blocker is on. I believe because it uses the window dataLayer object and does some background processes it doesnt affect the user

BriungRi (approved):

PR #252, opened 2024-04-07
made network name in WalletCore lowercased for consistent behaviour from different wallets

RowanFuture:
Petra Wallet made by Aptos Labs uses uppercase for the first letter of network names, so I'd say it's reasonable to accept this.

### DIFF
--- a/packages/wallet-adapter-core/src/WalletCore.ts
+++ b/packages/wallet-adapter-core/src/WalletCore.ts
@@ -1,5 +1,6 @@
 import { TxnBuilderTypes, Types, BCS } from "aptos";
 import {
+  Network,
   AnyRawTransaction,
   AccountAuthenticator,
   AccountAuthenticatorEd25519,
@@ -383,13 +384,13 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
     if (this._wallet?.isAIP62Standard) {
       const standardizeNetwork = network as StandardNetworkInfo;
       this._network = {
-        name: standardizeNetwork.name.toLowerCase(),
+        name: standardizeNetwork.name.toLowerCase() as Network,
         chainId: standardizeNetwork.chainId.toString(),
         url: standardizeNetwork.url,
       };
       return;
     }
-    this._network = { ...(network as NetworkInfo) };
+    this._network = { ...(network as NetworkInfo), name: network.name.toLowerCase() as Network };
   }
 
   /**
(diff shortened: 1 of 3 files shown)


0xmaayan:
This warning serves for a wallet to check whether it is compatible with what the adapter expects - which is a lower case network name.

RowanFuture:
> This warning serves for a wallet to check whether it is compatible with what the adapter expects - which is a lower case network name. 

So is Petra Wallet incompatible with the adapter?

0xmaayan:
> > This warning serves for a wallet to check whether it is compatible with what the adapter expects - which is a lower case network name.
> 
> So is Petra Wallet incompatible with the adapter?

I wont say incompatible as it definitely works, but it returns the network name different than expected. 

We can however handle it at a lower level in the adapter, and lowercase the network name as we set it https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-core/src/WalletCore.ts#L378

RowanFuture:
> > > This warning serves for a wallet to check whether it is compatible with what the adapter expects - which is a lower case network name.
> > 
> > 
> > So is Petra Wallet incompatible with the adapter?
> 
> I wont say incompatible as it definitely works, but it returns the network name different than expected.
> 
> We can however handle it at a lower level in the adapter, and lowercase the network name as we set it https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-core/src/WalletCore.ts#L378

Very good point,
I've updated the code.

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -383,7 +383,7 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
     if (this._wallet?.isAIP62Standard) {
       const standardizeNetwork = network as StandardNetworkInfo;
       this._network = {
-        name: standardizeNetwork.name,
+        name: standardizeNetwork.name.toLowerCase(),
```
0xmaayan: looks good. 
- You might want to also update it here https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-core/src/WalletCore.ts#L392 as we set the network name in 2 places
- Please also run `changeset` to get this change with the release of a new version https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/CONTRIBUTING.md#changesets

RowanFuture: Thanks! Both done.

0xmaayan (approved):

PR #257, opened 2024-04-18
Demo integrating T Wallet AIP-62 adapter

lcs86-dev:
This pull request shows the integration process of T Wallet AIP-62 adapter.

[Link to Integration Guide](https://atomrigs-1.gitbook.io/dekey-wallet/aptos-aip-62/quick-start)

Additional Notes:
- WalletCore might need to support secp256k1 and SingleKey for verifySignature in the signMessageAndVerify function.
- Registering an email with SKT Company is necessary to utilize the KYC feature to test this PR.
- Ensure that the demo DApp is available on localhost:3000 for local testing, as T Wallet's development system currently adds this domain for testing purposes.

https://github.com/aptos-labs/aptos-wallet-adapter/assets/85222860/54b9995e-ad67-4a87-81ed-f90db9253c12

### DIFF
--- a/apps/nextjs-example/package.json
+++ b/apps/nextjs-example/package.json
@@ -17,6 +17,7 @@
     "@aptos-labs/wallet-adapter-mui-design": "workspace:*",
     "@aptos-labs/wallet-adapter-react": "workspace:*",
     "@aptos-labs/wallet-standard": "^0.0.11",
+    "@atomrigslab/aptos-wallet-adapter": "^0.1.5",
     "@bitget-wallet/aptos-wallet-adapter": "^0.1.0",
     "@blocto/aptos-wallet-adapter-plugin": "^0.2.2",
     "@flipperplatform/wallet-adapter-plugin": "^0.1.0",
--- a/apps/nextjs-example/package.json
+++ b/apps/nextjs-example/package.json
@@ -17,7 +17,6 @@
     "@aptos-labs/wallet-adapter-mui-design": "workspace:*",
     "@aptos-labs/wallet-adapter-react": "workspace:*",
     "@aptos-labs/wallet-standard": "^0.0.11",
-    "@atomrigslab/aptos-wallet-adapter": "^0.1.5",
     "@bitget-wallet/aptos-wallet-adapter": "^0.1.0",
     "@blocto/aptos-wallet-adapter-plugin": "^0.2.2",
     "@flipperplatform/wallet-adapter-plugin": "^0.1.0",
--- a/apps/nextjs-example/pages/_app.tsx
+++ b/apps/nextjs-example/pages/_app.tsx
@@ -1,5 +1,8 @@
 import type { AppProps } from "next/app";
 import { AppContext } from "../components/AppContext";
+import { DevTWallet } from "@atomrigslab/aptos-wallet-adapter"
+
+new DevTWallet();
 
 // order matters
 import "../styles/global.css";
--- a/apps/nextjs-example/pages/_app.tsx
+++ b/apps/nextjs-example/pages/_app.tsx
@@ -1,8 +1,5 @@
 import type { AppProps } from "next/app";
 import { AppContext } from "../components/AppContext";
-import { DevTWallet } from "@atomrigslab/aptos-wallet-adapter"
-
-new DevTWallet();
 
 // order matters
 import "../styles/global.css";
--- a/packages/wallet-adapter-core/package.json
+++ b/packages/wallet-adapter-core/package.json
@@ -49,6 +49,7 @@
   },
   "dependencies": {
     "@aptos-labs/wallet-standard": "^0.0.11",
+    "@atomrigslab/aptos-wallet-adapter": "^0.1.5",
     "buffer": "^6.0.3",
     "eventemitter3": "^4.0.7",
     "tweetnacl": "^1.0.3"
--- a/packages/wallet-adapter-core/package.json
+++ b/packages/wallet-adapter-core/package.json
@@ -49,7 +49,7 @@
   },
   "dependencies": {
     "@aptos-labs/wallet-standard": "^0.0.11",
-    "@atomrigslab/aptos-wallet-adapter": "^0.1.6",
+    "@atomrigslab/aptos-wallet-adapter": "^0.1.10",
     "buffer": "^6.0.3",
     "eventemitter3": "^4.0.7",
     "tweetnacl": "^1.0.3"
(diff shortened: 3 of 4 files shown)


apps/nextjs-example/package.json
```diff
@@ -17,6 +17,7 @@
     "@aptos-labs/wallet-adapter-mui-design": "workspace:*",
     "@aptos-labs/wallet-adapter-react": "workspace:*",
     "@aptos-labs/wallet-standard": "^0.0.11",
+    "@atomrigslab/aptos-wallet-adapter": "^0.1.5",
```
0xmaayan: this should live in the core package, I can help with that

0xmaayan (changes requested):
Blocking land to do some tests and provide feedback as well as integrate the npm package in the adapter core level

apps/nextjs-example/package.json
lcs86-dev: > this should live in the core package, I can help with that

Thank you for the guide and help.

Should we install TWallet adapter in the core library dependency and set it up like in [this one](https://github.com/aptos-labs/aptos-wallet-adapter/blob/sdk_standard_wallets/packages/wallet-adapter-core/src/AIP62StandardWallets/sdkWallets.ts)?

Our adapter includes both DevTWallet and ProdTWallet. I intended to allow dApp developers to choose which one to use during development and production. While installing the dependency in the adapter is okay, enabling dapp to register TWallet might be better, as it lets dApps decide which one to use.

0xmaayan: > > this should live in the core package, I can help with that
> 
> Thank you for the guide and help.
> 
> Should we install TWallet adapter in the core library dependency and set it up like in [this one](https://github.com/aptos-labs/aptos-wallet-adapter/blob/sdk_standard_wallets/packages/wallet-adapter-core/src/AIP62StandardWallets/sdkWallets.ts)?
> 
> Our adapter includes both DevTWallet and ProdTWallet. I intended to allow dApp developers to choose which one to use during development and production. While installing the dependency in the adapter is okay, enabling dapp to register TWallet might be better, as it lets dApps decide which one to use.

SDK wallets should be installed and managed in the adapter core package.

That way, a wallet benefits from visibility, custom UI, adapter extra features and logic while the dapp utilizes the adapter for more security, state management and features.

Also, having a dapp uses the wallet adapter and a separate sdk wallet npm package can create state management risks and a buggy UI and the wallet/dapp will not utilize the adapter for all the extra features and logic.

Yes the process of adding a SDK wallet should look similar to [that](https://github.com/aptos-labs/aptos-wallet-adapter/blob/sdk_standard_wallets/packages/wallet-adapter-core/src/AIP62StandardWallets/sdkWallets.ts), tho need some more configuration - I can help with that when I am back next week.

0xmaayan:
Was able to test the initial integration using this [branch](https://github.com/aptos-labs/aptos-wallet-adapter/blob/sdk_standard_wallets/packages/wallet-adapter-core/src/AIP62StandardWallets/sdkWallets.ts) and have some feedbacks

- Importing the "DevTWallet" in the adapter, i.e `import { DevTWallet } from "@atomrigslab/aptos-wallet-adapter";`, it looks like it initiates both dev and prod wallet classes (and we end up with 2 T wallets on the UI) 
Screenshot 2024-04-29 at 12 54 06 PM

- Importing "TWallet",i.e `import { TWallet } from "@atomrigslab/aptos-wallet-adapter";`, initiates only the prod wallet class
Screenshot 2024-04-29 at 12 52 48 PM

- Connecting with "T Wallet", returns an error (you can see it on the demo video you linked here and could also reproduce it on my local)

lcs86-dev:
Thank you for the feedback.

I suspect that the issue during local development, where using HMR with the local development server might result in two TWallets on the UI. When using new DevTWallet() or new ProdTWallet(), only one registration event is emitted.

Attempting to connect to TWallet currently triggers an error as the production TWallet is not deployed yet. It's under the QA process and will be published soon.

0xmaayan:
Thanks!

Could you clarify? Is this an expected behavior during local development?

> I suspect that the issue during local development, where using HMR with the local development server might result in two TWallets on the UI. When using new DevTWallet() or new ProdTWallet(), only one registration event is emitted.

lcs86-dev:
Yes. When a new DevTWallet or TWallet is created, they emit a registration event. It seems that the application retains the information of the previously registered wallet.

new DevTwallet() or new TWallet() internally calls the registerWallet function from @aptos-labs/wallet-standard only once.

If both DevTwallet and TWallet buttons appear, dApp page might need to be refreshed and it works as expected.

0xmaayan:
> Yes. When a new DevTWallet or TWallet is created, they emit a registration event. It seems that the application retains the information of the previously registered wallet.
> 
> new DevTwallet() or new TWallet() internally calls the registerWallet function from @aptos-labs/wallet-standard only once.
> 
> If both DevTwallet and TWallet buttons appear, dApp page might need to be refreshed and it works as expected.

I am not completely sure why that would be an expected behavior on local dev.

Take a look at [this](https://github.com/aptos-labs/aptos-wallet-adapter/tree/sdk_standard_wallets/packages/wallet-adapter-core) branch

[here](https://github.com/aptos-labs/aptos-wallet-adapter/blob/sdk_standard_wallets/packages/wallet-adapter-core/src/AIP62StandardWallets/sdkWallets.ts) I import the `DevTWallet`
then I [initiate](https://github.com/aptos-labs/aptos-wallet-adapter/blob/sdk_standard_wallets/packages/wallet-adapter-core/src/WalletCore.ts#L182) it and add the class to the wallets array, and finally both wallets show up on the UI on first load and every time I refresh the page.
image

There might be some wrong logic in the T wallet SDK that initiates/calls both wallet classes.

By the way, if I change the [import](https://github.com/aptos-labs/aptos-wallet-adapter/blob/sdk_standard_wallets/packages/wallet-adapter-core/src/AIP62StandardWallets/sdkWallets.ts#L1) to `TWallet` then only one wallet is shown.

lcs86-dev:
Apologies for not catching this bug earlier. As you mentioned, the previous version (0.1.5) of the adapter unintentionally registered TWallet instances when creating a DevTWallet.

I've made updates to the code to prevent triggering the TWallet registration event when creating a DevTWallet or any other environment-specific TWallet. This issue has been addressed in version 0.1.6. Thanks for pointing this out.

0xmaayan:
@lcs86-dev Thanks for the PR. I moved T Wallet integration effort to this PR https://github.com/aptos-labs/aptos-wallet-adapter/pull/279 as the adapter needs internal logic implementation to support AIP-62 compatible sdk wallets.

0xmaayan:
closing as it is complete

PR #261, opened 2024-04-29
Support a custom network defined in a Wallet

0xmaayan:
Some Wallets support the option to provide a custom RPC to work with,  this PR adds support for custom networks defined in a wallet to be respected by the Wallet Adapter.

If the current connected wallet configured to work with a non Aptos network, the adapter resolves the custom network to use when submitting a transaction

### DIFF
--- a/apps/nextjs-example/pages/index.tsx
+++ b/apps/nextjs-example/pages/index.tsx
@@ -32,12 +32,12 @@ const WalletSelectorAntDesign = dynamic(
   }
 );
 
-const isSendableNetwork = (connected: boolean, network?: string): boolean => {
-  return connected && !isMainnet(connected, network);
+const isSendableNetwork = (connected: boolean): boolean => {
+  return connected && !isMainnet(connected);
 };
 
-const isMainnet = (connected: boolean, network?: string): boolean => {
-  return connected && network?.toLowerCase() === Network.MAINNET.toLowerCase();
+const isMainnet = (connected: boolean, networkName?: string): boolean => {
+  return connected && networkName === Network.MAINNET;
 };
 
 export default function App() {
--- a/apps/nextjs-example/pages/index.tsx
+++ b/apps/nextjs-example/pages/index.tsx
@@ -1,4 +1,9 @@
-import { useWallet, AccountInfo, NetworkInfo, WalletInfo} from "@aptos-labs/wallet-adapter-react";
+import {
+  useWallet,
+  AccountInfo,
+  NetworkInfo,
+  WalletInfo,
+} from "@aptos-labs/wallet-adapter-react";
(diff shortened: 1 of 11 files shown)


packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -48,6 +48,8 @@ import {
   scopePollingDetectionStrategy,
   isRedirectable,
   generalizedErrorMessage,
+  getAptosConfig,
+  isAptosNetwork,
```
davidiw, line 48: how come the linter doesn't organize these?

davidiw (approved):
curious why we don't sanitize the network name in one space and then reimport from there?

apps/nextjs-example/pages/index.tsx
```diff
@@ -27,11 +33,11 @@ const WalletSelectorAntDesign = dynamic(
 );
 
 const isSendableNetwork = (connected: boolean, network?: string): boolean => {
```
BriungRi: thoughts on doing something like `Lowercase<string>` instead of `string`? That way, lowercase is enforced at type level

https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html#lowercasestringtype

BriungRi (approved):

packages/wallet-adapter-core/src/utils/helpers.ts
```diff
@@ -52,3 +58,41 @@ export const areBCSArguments = (
...
+ * @param networkInfo
+ * @returns boolean
+ */
+export const isAptosNetwork = (
+  networkInfo: NetworkInfo | StandardNetworkInfo | null
+): boolean => {
+  if (!networkInfo) {
+    throw new Error("Undefined network");
+  }
+  return NetworkToNodeAPI[networkInfo.name] ? true : false;
```
BriungRi: simplify as `return NetworkToNodeAPI[networkInfo.name] !== undefined`?

0xmaayan:
> curious why we don't sanitize the network name in one space and then reimport from there?

Actually, we do https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-core/src/WalletCore.ts#L406-L420. The demo dapp needs to be updated to not cast it.

packages/wallet-adapter-core/src/WalletCore.ts
0xmaayan: It doesnt have this rule

0xmaayan:
> > curious why we don't sanitize the network name in one space and then reimport from there?
> 
> Actually, we do https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-core/src/WalletCore.ts#L406-L420. The demo dapp needs to be updated to not cast it.

updated

PR #268, opened 2024-05-01
Block scam transaction submission

0xmaayan:
Block calling `0x1::account::rotate_authentication_key_call` entry function

### DIFF
--- a/packages/wallet-adapter-core/src/WalletCore.ts
+++ b/packages/wallet-adapter-core/src/WalletCore.ts
@@ -633,6 +633,16 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
     { hash: Types.HexEncodedBytes; output?: any } | PendingTransactionResponse
   > {
     try {
+      if ("function" in transactionInput.data) {
+        if (
+          transactionInput.data.function ===
+          "0x1::account::rotate_authentication_key_call"
+        ) {
+          throw new WalletSignAndSubmitMessageError("SCAM SITE DETECTED")
+            .message;
+        }
+      }
+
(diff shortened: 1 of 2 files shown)


packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -626,6 +627,16 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
     try {
+      if ("function" in transactionInput.data) {
+        if (
+          transactionInput.data.function ===
+          "0x1::account::rotate_authentication_key_call"
+        ) {
+          throw new WalletSignAndSubmitMessageError("SCAM SITE DETECTED")
+            .message;
+        }
+      }
```
gregnazario, line 644: It's possible that this could be used for other purposes.  But, let's go with this for now, and let wallets handle rotation directly.

gregnazario (approved):

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -633,6 +633,16 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
     { hash: Types.HexEncodedBytes; output?: any } | PendingTransactionResponse
   > {
     try {
+      if ("function" in transactionInput.data) {
+        if (
+          transactionInput.data.function ===
+          "0x1::account::rotate_authentication_key_call"
+        ) {
+          throw new WalletSignAndSubmitMessageError("SCAM SITE DETECTED")
+            .message;
```
hardsetting, line 642: Why not throwing the actual error object?

PR #272, opened 2024-05-08
Add support for uninstalled AIP-62 compatible wallets

0xmaayan:
With the new wallet standard (AIP-62) it is hard to display extension wallets that are not installed on the user machine. 

This PR introduces a registry for AIP-62 compatible wallets so the adapter could display these wallets to the end user on a wallet selector modal and provide the option to install these wallets.

Additionally, the adapter detects the device type and modifies the wallet selector modal list based on device type
- remove wallets not compatible with device

See https://aptos-labs.github.io/aptos-wallet-adapter/nextjs-example-testing/ - can try it on a browser where Nightly is not installed (as Nightly is the only AIP-62 compatible wallet these days)

On a mobile view - only mobile compatible wallets are shown
IMG_7667


packages/wallet-adapter-core/src/AIP62StandardWallets/registry.ts
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * AIP-62 compatible wallets are required to add their wallet info here if they want to be detected by the adapter
+ *
+ * name - The wallet name
+ * url - The wallet website url (where users can download it from)
+ * icon - The wallet icon
+ */
+export const aptosStandardSupportedWalletList: Array<AptosStandardSupportedWallet> =
+  [
+    {
```
gregnazario, line 17: Do we want to possibly look at making this decentralized in the future?  Or are we just hoping everyone contributes here.

(Or even let people customize which ones show up)

gregnazario (approved):

packages/wallet-adapter-core/src/AIP62StandardWallets/registry.ts
0xmaayan: Yea, we would probably modify it a bit in the future, for now we only have Nightly and we want it to show up for users so we go with this initial implementation

PR #273, opened 2024-05-09
Fix the `useEffect` dependency in auto-connect

xbtmatt:
### Description

The `useEffect` dependency in the auto-connect block in `WalletProvider.tsx` was using the `wallets` array directly rather than as a single dependency in the dependency array, which seems like it was most likely a typo.

This wasn't an issue before because the `wallets` value was static and never changed upon load. Now that AIP-62 compliant wallets load at a separate time than the non-AIP-62 wallets, this value changes and thus React now warns you:

```shell
client.js:2 Warning: The final argument passed to useEffect changed size between renders. The order and size of this array must remain constant.

Previous: [[object Object], [object Object], [object Object], [object Object]]
Incoming: [[object Object], [object Object], [object Object], [object Object], [object Object]]
```

Note the additional object in the `Incoming` array is the `Nightly` wallet

I've updated it to be correct:

```tsx
  useEffect(() => {
    if (autoConnect) {
      if (localStorage.getItem("AptosWalletName") && !connected) {
        connect(localStorage.getItem("AptosWalletName") as WalletName);
      } else {
        // if we dont use autoconnect set the connect is loading to false
        setIsLoading(false);
      }
    }
  }, [wallets]);  //  <-------------- was `wallets` before
  ```
  
I also fixed a few typos in the repo 😄

### DIFF
--- a/.changeset/odd-dingos-lie.md
+++ b/.changeset/odd-dingos-lie.md
@@ -1,0 +1,5 @@
+---
+"@aptos-labs/wallet-adapter-react": patch
+---
+
+Updated the `useEffect` dependency array for auto-connecting
--- a/.changeset/odd-dingos-lie.md
+++ b/.changeset/odd-dingos-lie.md
@@ -2,4 +2,4 @@
 "@aptos-labs/wallet-adapter-react": patch
 ---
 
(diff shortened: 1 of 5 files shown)


0xmaayan (approved):
❤️

PR #275, opened 2024-05-10
[deps] Upgrade NextJS for example

gregnazario:


0xmaayan:
Have we tested the demo dapp doesnt break?

gregnazario:
> Have we tested the demo dapp doesnt break?

I'm testing it now, since the build is broken :(

PR #277, opened 2024-05-13
Exposing wallet adapter core into window object

hardsetting:
In order to enable dynamic behaviors, we're now exposing the dapp's wallet adapter version through the window object, when available


bv-aptos (approved):
It's beautiful!

Ship it!

0xmaayan:
Dont manually upgrade version, but use pnpm run changeset

PR #279, opened 2024-05-14
Add support for AIP-62 compatible SDK wallets

0xmaayan:
Following AIP-62, SDK wallets expected to be implemented within the adapter core so the wallet adapter can manage and handle their packages.

This PR adds `T Wallet` as an SDK wallet and implements a logic to include it in the available wallets array.

### DIFF
--- a/packages/wallet-adapter-core/src/WalletCore.ts
+++ b/packages/wallet-adapter-core/src/WalletCore.ts
@@ -24,8 +24,10 @@ import {
   NetworkInfo as StandardNetworkInfo,
   UserResponse,
   UserResponseStatus,
+  isWalletWithRequiredFeatureSet,
 } from "@aptos-labs/wallet-standard";
 
+import SDKWallets from "./AIP62StandardWallets/sdkWallets";
 import { ChainIdToAnsSupportedNetworkMap, WalletReadyState } from "./constants";
 import {
   WalletAccountChangeError,
@@ -132,6 +134,8 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
     this.fetchAptosWallets();
     // Append AIP-62 compatible wallets that are not detected on the user machine
     this.appendNotDetectedStandardSupportedWallets(this._standard_wallets);
+    // Stretegy to detect AIP-62 standard compatible SDK wallets
+    this.fetchSDKWallets();
   }
 
   private scopePollingDetectionStrategy() {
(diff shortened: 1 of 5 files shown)


packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -194,7 +198,23 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
+
+    SDKWallets.map((sdkWallet: AptosStandardWallet) => {
+      const isValid = isWalletWithRequiredFeatureSet(sdkWallet);
+      // TODO add user opt-in check
+      if (isValid) {
+        const standardWallet = sdkWallet;
+        standardWallet.readyState = WalletReadyState.Installed;
+        aptosStandardWallets.push(standardWallet);
+        this.standardizeStandardWalletToPluginWalletType(standardWallet);
+      }
```
hardsetting: A little confused here.
Why do we have to cast, and manually set readyState?

hardsetting: Calling `setWallets` would override the `_standard_wallets` value right?

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -194,15 +196,28 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
+   *
+   * Loop over local SDK and Extensions wallets
+   * 1) check it is Standard compatible
+   * 2) Update their readyState to Installed (for a future UI detection)
+   * 3) push the wallet into a local wallets array
+   * 4) standardize each wallet to the Wallet Plugin type interface for legacy compatibility
+   *
+   * @param extensionwWallets
+   */
+  private setWallets(extensionwWallets: readonly AptosWallet[]) {
```
0xmaayan, line 210: @hardsetting please take another look, I have updated the code.

Can see a live example here https://aptos-labs.github.io/aptos-wallet-adapter/nextjs-example-testing/

hardsetting (approved):

packages/wallet-adapter-core/src/WalletCore.ts
hardsetting: Just checking that `this.fetchAptosWallets()` is always called so that `this._aptos_wallets` is always initialized

0xmaayan: yes, it is part of the `WalletCore` constructor method, so always called
https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-core/src/WalletCore.ts#L132

PR #283, opened 2024-05-21
Wallets opt-in support

0xmaayan:
Add support for the user to opt-in to specific wallets they want to include in their dapp.

This logic is only for AIP-62 compatible wallets.

```
 <AptosWalletAdapterProvider
      plugins={wallets}
      optInWallets=["Nightly","T wallet"] // include Nightly and T wallet wallets
    >
  {children}
</AptosWalletAdapterProvider>
```

### DIFF
--- a/packages/wallet-adapter-core/src/AIP62StandardWallets/types.ts
+++ b/packages/wallet-adapter-core/src/AIP62StandardWallets/types.ts
@@ -8,3 +8,5 @@ export interface AptosStandardSupportedWallet<Name extends string = string> {
   readyState: WalletReadyState.NotDetected;
   isAIP62Standard: true;
 }
+
+export type AvailableWallets = "Nightly" | "T wallet";
--- a/packages/wallet-adapter-core/src/WalletCore.ts
+++ b/packages/wallet-adapter-core/src/WalletCore.ts
@@ -242,7 +242,7 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
    * @param walletName
    * @returns
    */
-  private excludeWallet(walletName: string): boolean {
+  excludeWallet(walletName: string): boolean {
     // If _optInWallets is not empty, and does not include the provided wallet,
     // return true to exclude the wallet, otherwise return false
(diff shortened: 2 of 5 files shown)


packages/wallet-adapter-core/src/AIP62StandardWallets/types.ts
```diff
@@ -8,3 +8,5 @@ export interface AptosStandardSupportedWallet<Name extends string = string> {
   readyState: WalletReadyState.NotDetected;
   isAIP62Standard: true;
 }
+
+export type AvailableWallets = "Nightly" | "T wallet";
```
hardsetting, line 12: Do we have a list of default wallets we can obtain this list from?

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -221,6 +232,26 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
+  private excludeWallet(walletName: string): boolean {
+    // If _optInWallets is not empty, and does not include the provided wallet,
+    // return true to exclude the wallet, otherwise return false
+    if (
+      this._optInWallets.length > 0 &&
+      !this._optInWallets.includes(walletName as AvailableWallets)
+    ) {
+      return true;
+    }
+    return false;
```
hardsetting: nit: not sure if this needs to be a private method.
It's only used once and it's just simple filtering that most developer understand intuitively.

```typescript
// Filter out wallets when the user has explicitly provided an opt-in list
if (this._optInWallets.length > 0 && !this_optInWallets.includes(walletName) {
  return;
}
```

hardsetting (approved):
lgtm!

packages/wallet-adapter-core/src/AIP62StandardWallets/types.ts
0xmaayan: not really. we have a wallet registry list only for extension wallets. This can be the list 😄

PR #287, opened 2024-05-24
Support account prop to be of AIP-62 `AccountInfo` type

0xmaayan:
The adapter returns the current connected wallet account address and public key as a `string` type. With the latest AIP-62 upgrade, the Wallet now returns the account address as a `AccountAddress` type and the public ket as `PublicKey` type.

This change is to support the 2 new types to be returned correctly from the adapter.
With this change, the `account` type returned by the adapter is of type [AccountInfo](https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-core/src/LegacyWalletPlugins/types.ts#L56) | [NewStandardAccountInfo](https://github.com/aptos-labs/wallet-standard/blob/main/src/AccountInfo.ts#L17)

Before
```
const {account} = useWallet();
account.address // string type
account.publicKey // string type
```

Now
```
const {account} = useWallet();
account.address // string | AccountAddress type
account.publicKey // string | PublicKey type
```

### DIFF
--- a/apps/nextjs-example/pages/index.tsx
+++ b/apps/nextjs-example/pages/index.tsx
@@ -13,11 +13,11 @@ import SponsorTransaction from "../components/transactionFlow/Sponsor";
 import MultiAgentTransaction from "../components/transactionFlow/MultiAgent";
 import Row from "../components/Row";
 import Col from "../components/Col";
-import { Network } from "@aptos-labs/ts-sdk";
-import { Select, Typography } from "antd";
+import { AccountAddress, Network } from "@aptos-labs/ts-sdk";
+import { Typography } from "antd";
 import {
   AptosChangeNetworkOutput,
-  StandardNetworkInfo,
+  NewStandardAccountInfo,
   isAptosNetwork,
 } from "@aptos-labs/wallet-adapter-core";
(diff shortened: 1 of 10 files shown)


gregnazario (commented):
so, if people continue provide a string version, then you won't know the type of key?

Should we provide an optional key type?  There's not going to be any loss of type conversion across this boundary is there?

0xmaayan:
> so, if people continue provide a string version, then you won't know the type of key?
> 
> Should we provide an optional key type? There's not going to be any loss of type conversion across this boundary is there?

It should come from the wallet side, the new standard requires the wallet to provide a typed public key

gregnazario (approved):

apps/nextjs-example/pages/index.tsx
```diff
@@ -47,7 +48,7 @@ const isMainnet = (connected: boolean, networkName?: string): boolean => {
 
 export default function App() {
   const { account, connected, network, wallet, changeNetwork } = useWallet();
-
+  console.log("account", account);
```
gregnazario: Oh probably remove this?

PR #297, opened 2024-05-30
Add AptosConnect

0xmaayan:
- Update SDK to version 1.18.1 to include keyless
- Use AptosConnect plugin version 0.0.11

### DIFF
--- a/apps/nextjs-example/components/transactionFlow/MultiAgent.tsx
+++ b/apps/nextjs-example/components/transactionFlow/MultiAgent.tsx
@@ -2,6 +2,7 @@ import {
   Account,
   AccountAuthenticator,
   AnyRawTransaction,
+  Ed25519Account,
 } from "@aptos-labs/ts-sdk";
 import { useWallet } from "@aptos-labs/wallet-adapter-react";
 
@@ -23,7 +24,7 @@ export default function MultiAgentTransaction({
     useWallet();
 
   const [secondarySignerAccount, setSecondarySignerAccount] =
-    useState<Account>();
+    useState<Ed25519Account>();
   const [transactionToSubmit, setTransactionToSubmit] =
     useState<AnyRawTransaction | null>(null);
   const { setSuccessAlertHash } = useAlert();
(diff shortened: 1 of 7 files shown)


0xmaayan:
getting this error when trying to connect AptosConnect on the adapter demo dapp

https://github.com/aptos-labs/aptos-wallet-adapter/assets/29798064/3aba3404-4672-476f-97db-7ae59b20b559

packages/wallet-adapter-core/src/AIP62StandardWallets/sdkWallets.ts
```diff
@@ -1,10 +1,25 @@
...
+  // a newer version (0.1.0) - this causes type mismatch. We should figure out how to handle it.
+  sdkWallets.push(new TWallet() as any);
+} else {
+  sdkWallets.push(new DevTWallet() as any);
+}
+
+let aptosNetwork = isProd ? Network.MAINNET : Network.TESTNET;
+// Need to check window is defined for AptosConnect
+if (typeof window !== "undefined") {
+  sdkWallets.push(new AptosConnectWallet({ network: aptosNetwork }));
```
hardsetting, line 22: what if the dapp developer wants to use a different config?
e.g. different network, frontendBaseURL or a dappId (to support IC accounts)

Ideally, only if the dapp developer has not registered AptosConnect, then we initialize it as done above.
Might also need to take into consideration race conditions if we have dedupe logic

hardsetting (approved):

apps/nextjs-example/components/transactionFlow/MultiAgent.tsx
```diff
@@ -23,7 +24,7 @@ export default function MultiAgentTransaction({
     useWallet();
 
   const [secondarySignerAccount, setSecondarySignerAccount] =
-    useState<Account>();
+    useState<Ed25519Account>();
```
hardsetting, line 27: Is this change relevant to the PR?

alyssaponzo (approved):

apps/nextjs-example/components/transactionFlow/MultiAgent.tsx
0xmaayan: yes, newest sdk version removed the privateKey from Account

packages/wallet-adapter-core/src/AIP62StandardWallets/sdkWallets.ts
0xmaayan: got it. we can work on it in a next PR

PR #298, opened 2024-05-31
Add the `local` case to the network string conversion function and narrow `Account` type to `Ed25519Account` in example

xbtmatt:
Currently you cannot use the wallet adapter in a local development environment with certain wallets because the middleware network string conversion fails, since it doesn't include the `local` case.

- Added the `local` case to the switch statement that converts a string to a Network.


apps/nextjs-example/components/transactionFlow/MultiAgent.tsx
```diff
@@ -43,7 +43,7 @@ export default function MultiAgentTransaction({
       throw new Error("no network");
     }
 
-    const secondarySigner = Account.generate();
+    const secondarySigner = Ed25519Account.generate();
```
0xmaayan: nit:

dont really need this change as `Account.generate()` already creates a legacy Ed25519Account

apps/nextjs-example/components/transactionFlow/MultiAgent.tsx
```diff
@@ -23,7 +23,7 @@ export default function MultiAgentTransaction({
     useWallet();
 
   const [secondarySignerAccount, setSecondarySignerAccount] =
-    useState<Account>();
+    useState<Ed25519Account>();
```
0xmaayan: this is definitely a needed change once the adapter is updated to use latest sdk version . check out https://github.com/aptos-labs/aptos-wallet-adapter/pull/296/files 😄

0xmaayan (commented):
let's remove the Account change in favor of the other PR that updates the SDK and holds this change already

apps/nextjs-example/components/transactionFlow/MultiAgent.tsx
xbtmatt: Ah ok, I only did it to make it more explicit :p, wasn't sure if it'd type it correctly

0xmaayan (approved):

PR #310, opened 2024-06-07
Add headless components and shadcn example

blakezimmerman:
This PR adds some headless components and utility functions to `@aptos-labs/wallet-adapter-react` to make it easier to build custom wallet connect UIs. This PR also overhauls the wallet adapter demo dapp by rewriting it with Next.js v14 and shadcn/ui (bringing mobile-responsiveness and dark mode). The shadcn/ui wallet selector serves as an example for using the newly added headless components and utils.

https://github.com/aptos-labs/aptos-wallet-adapter/assets/25761639/bdd67ffa-0fd5-4076-8135-4e448d5623a3


packages/wallet-adapter-react/src/utils.tsx
```diff
@@ -0,0 +1,45 @@
+import {
+  AptosStandardSupportedWallet,
+  Wallet,
+  WalletReadyState,
+  isRedirectable,
+} from "@aptos-labs/wallet-adapter-core";
+
+export type AptosWallet = Wallet | AptosStandardSupportedWallet<string>;
```
0xmaayan: that could live on the core package? and then any package can import it

apps/shadcn-example/src/components/wallet-connection/connect-wallet.tsx
```diff
@@ -0,0 +1,78 @@
+import {
```
0xmaayan: any reason we dont simply include it in the main demo dapp?

apps/shadcn-example/src/components/wallet-connection/connect-wallet.tsx
```diff
@@ -0,0 +1,78 @@
...
+  AptosWallet,
+  WalletItem,
+  isInstallRequired,
```
0xmaayan: what is the end dev experience when using the shadcn selector modal? I would expect the user dont need to do everything we see here on the example dapp.

I can imagine something like, on the dapp side,
```
import {WalletSelector} from "@aptos-labs/wallet-adapter-react"
```
and then WalletSelector just holds everything (same as current antd), with probably properties we can provide to it.

Is this what you expect? or you have something else in mind?

packages/wallet-adapter-react/src/utils.tsx
blakezimmerman: That's a good idea, I'll move it there

apps/shadcn-example/src/components/wallet-connection/connect-wallet.tsx
blakezimmerman: So right now, if you're using shadcn, you would pretty much just copy all the code from this directory into your app and ensure you install all the shadcn components it makes use of. Ideally, we would have an `@aptos-labs/wallet-adapter-shadcn` package that handles that for you, but since shadcn is a CLI and not an actual NPM package, I would have to think more about how that would work.

The goal of the `WalletItem` headless component and its associated utility functions is to make creating custom wallet selectors easier without asserting any design decisions. It's still a very low-level abstraction and there's a lot more we can build on top of it. The benefit is that you can reuse your own dialog implementation or choose not to use a dialog at all (for example, Find Out had a wallet connect page instead of a dialog).

I'll try to make this more clear in the documentation, but these headless components are building blocks for making a wallet adapter with your own design system rather than a single component you just drop in your app. In the future, we can use these building blocks to quickly create more packages like maybe an `@aptos-labs/wallet-adapter-ark-ui`, which would be closer to a single component that you drop into your app since Ark UI provides dialog, collapsible, and dropdown menu components.

0xmaayan: any reason not getting the example code as is into a@aptos-labs npm package, and have the end user simply import the WalletItem & co? there is a bigger vision to it and it sounds interesting but I think meanwhile we can wrap it all in a package to ease the end dev experience (without expecting the end dev to add more shadcn component to it, etc)

blakezimmerman: I thought about this a bit over the weekend, and unfortunately I don't think it would make sense to wrap this shadcn wallet selector in a package. At the start of the [shadcn docs](https://ui.shadcn.com/docs#:~:text=This%20is%20NOT%20a%20component%20library.%20It%27s%20a%20collection%20of%20re%2Dusable%20components%20that%20you%20can%20copy%20and%20paste%20into%20your%20apps.), they state:

> This is NOT a component library. It's a collection of re-usable components that you can copy and paste into your apps.

Copy/Paste and code ownership are the core philosophies of shadcn. I was looking at 3rd party shadcn-based components and they all basically say "just copy this file to your components directory". Shadcn isn't like most UI tools, and serves more of a starting point for your UI rather than a be-all and end-all solution.

To make the process simpler, I put the `WalletSelector` component in a single ~150 line `wallet-selector.tsx` that users can copy into their app. Also, I updated the `README.md` of this app with step-by-step instructions. In those instructions I provide a single command they can run to ensure they have all the prerequisite components:

```bash
npx shadcn-ui@latest add button collapsible dialog dropdown-menu toast
```

If you think about these wallet adapter UIs as a sliding scale from simplicity to customizability, our MUI and AntD packages are at the far end of simplicity and the solution in this PR is at the far end of customizability. If you just want to drop a component into your app and call it a day, the MUI and AntD packages are still the go-to solutions. If you want to roll your own wallet selector with your app's styles and existing dialog component, then the `WalletItem` headless component and new utility functions will greatly simplify that.

With all that being said, once I finish all of the headless components for handling Aptos Connect, I do plan on adding a new package to this repo that sits at the middle of the scale between simplicity and customizability. It will export a single `WalletSelector` component and include it's own implementation of a dialog, dropdown menu, collapsible, etc. The goal is to eventually build up this mildly opinionated package to something like [Family ConnectKit](https://docs.family.co/connectkit/theming).

packages/wallet-adapter-react/src/utils.tsx
blakezimmerman: I just realized that `@aptos-labs/wallet-standard` already exports an interface called `AptosWallet`. Any idea what we should call this instead?

0xmaayan: `AnyAptosWallet`? 😄

apps/shadcn-example/src/components/wallet-selector.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+import {
+  AptosWallet,
+  WalletItem,
+  isInstallRequired,
+  partitionWallets,
+  truncateAddress,
+  useWallet,
+} from "@aptos-labs/wallet-adapter-react";
+import { ChevronDown, Copy, LogOut } from "lucide-react";
```
kaw2k: Is this a package included in the shadcn installation?

apps/shadcn-example/src/components/wallet-connection/connect-wallet.tsx
kaw2k: This makes sense to me. Anyone already familiar with shadcn won't be daunted by these steps. In the absence of a CLI tool, perhaps in addition to linking to the file you could include a `curl` command a user could copy and paste. Just an idea.

apps/shadcn-example/.eslintrc.json
```diff
@@ -0,0 +1,4 @@
+{
```
0xmaayan: I think I had a comment on that before, but any reason we dont simply include it in the existing demo dapp instead of creating a new example?

apps/shadcn-example/src/components/wallet-selector.tsx
blakezimmerman: Yes, it's the default icon pack that's included with shadcn

packages/wallet-adapter-react/src/utils.tsx
blakezimmerman: That's good enough for me. I added that type to wallet-core

apps/shadcn-example/.eslintrc.json
blakezimmerman: I figured that having a standalone shadcn example would be easier to understand from a user standpoint if they just want to add a wallet selector to their own shadcn app. Also, I would have to make a decent amount of changes to the existing demo dapp to add the light/dark mode switcher, and I wasn't sure if adding tailwind and all the shadcn components would make the dapp seem too busy.

Let me know what you think though. I probably could still merge them together.

apps/shadcn-example/src/components/wallet-connection/connect-wallet.tsx
blakezimmerman: I replied to your later comment on this

apps/shadcn-example/.eslintrc.json
0xmaayan: that makes sense. the idea for the demo dapp is 1) show users how they can use the adapter and 2) have a live site users can go to and see how everything works https://aptos-labs.github.io/aptos-wallet-adapter/

We can either have it all in one site or maybe have 2 different demo sites (which seems to much imo)

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -83,6 +83,8 @@ import { aptosStandardSupportedWalletList } from "./AIP62StandardWallets/registr
 
 export type IAptosWallet = AptosStandardWallet & Wallet;
 
+export type AnyAptosWallet = Wallet | AptosStandardSupportedWallet;
```
0xmaayan, line 89: could we add a comment explaining what this type is for

packages/wallet-adapter-react/CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # @aptos-labs/wallet-adapter-react
 
+## 3.1.6
+
+### Minor Changes
+
+- cb8acd1: Added the following headless React component and some utility functions to simplify the creation of new wallet adapter UIs:
```
0xmaayan: have you added it manually? we shouldn't. 

the way we handle changelog and releases is with changeset  https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/CONTRIBUTING.md#changesets

packages/wallet-adapter-react/package.json
```diff
@@ -41,11 +41,12 @@
     "@types/react": "^18.0.17",
     "@types/react-dom": "^18.0.6",
     "eslint": "^8.15.0",
-    "typescript": "^4.5.3",
-    "tsup": "^5.10.1"
+    "tsup": "^5.10.1",
+    "typescript": "^4.5.3"
```
0xmaayan, line 45: need to revert, this whole things should be handled with changeset

packages/wallet-adapter-react/CHANGELOG.md
blakezimmerman: Oops, I'll actually read the contributing doc and do it the right way

apps/shadcn-example/.eslintrc.json
blakezimmerman: That sounds fair. I'll re-style the deployed example dapp with shadcn and then delete the one I created. The MUI and AntD buttons will still stay obviously.

packages/wallet-adapter-core/src/WalletCore.ts
blakezimmerman: Just added a JSDoc to the type

apps/shadcn-example/.eslintrc.json
blakezimmerman: I just pushed a commit to merge the two examples together.

packages/wallet-adapter-core/src/WalletCore.ts
0xmaayan: nit: this can live under a `type` file https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-core/src/AIP62StandardWallets/types.ts

0xmaayan (commented):
lgtm. just need changesets and it is ready to go

kent-white (approved):
i only skimmed through the wallet-adapter-react piece. That seems good to me

packages/wallet-adapter-react/CHANGELOG.md
blakezimmerman: The changesets have been added

0xmaayan (approved):
nice!

PR #315, opened 2024-06-13
Fix bug in shadcn/ui wallet selector

blakezimmerman:
Fixed a bug where "More wallets" would still be displayed in the shadcn/ui wallet selector even if there aren't more wallets.


gregnazario (approved):

0xmaayan (approved):
❤️

PR #326, opened 2024-06-16
Support an optional `aptosConnectConfig` user prop to set AptosConnect configuration

0xmaayan:


packages/wallet-adapter-react/src/WalletProvider.tsx
```diff
@@ -33,6 +33,7 @@ export interface AptosWalletProviderProps {
   plugins?: ReadonlyArray<Wallet>;
   optInWallets?: ReadonlyArray<AvailableWallets>;
   autoConnect?: boolean;
+  aptosConnectConfig?: { network: Network };
```
hardsetting: How about just calling the option "defaultNetwork", so we can reuse it for other sdk wallets if applicable / feels less hacky

packages/wallet-adapter-react/src/WalletProvider.tsx
```diff
@@ -54,6 +55,7 @@ const initialState: {
  * @param plugins Non AIP-62 supported wallet plugins array
  * @param optInWallets AIP-62 supported wallet names array to only include in the adapter wallets
  * @param autoConnect A boolean flag to indicate if the adapter should auto connect to a wallet
+ * @param aptosConnectConfig A JSON format configurations to connect with AptosConnect
```
hardsetting:
```suggestion
 * @param aptosConnectConfig configuration passed to the auto-injected AptosConnect plugin
```

also I wouldn't mention JSON if not in the context of serialization or transport

hardsetting (approved):
I still think we should allow for deduplication, but this works for now :)

0xmaayan:
> deduplication

could you elaborate? if you mean letting the dapp to pass in the wallet plugin in the plugins array - I highly recommend not to do it. From past experience it introduces so many issues when there is a version mismatch between the dapp<>adapter<>the wallet plugin

packages/wallet-adapter-react/src/WalletProvider.tsx
0xmaayan: yea can do. can maybe rename to `dappConfig:{network: Network}` to support potential future dapp config (like client_id)

hardsetting:
> could you elaborate?

From the dapp:

```typescript
import { registerAptosConnect } from '@aptos-connect/wallet-adapter-plugin';

registerAptosConnect({ network: Network.TESTNET });

// later when initializing the wallet adapter, it detects AptosConnect was already registered and won't try to automatically inject it
...
```

> From past experience it introduces so many issues when there is a version mismatch between the dapp<>adapter<>the wallet plugin

Yeah that can easily get messy but it's a bit better with the new wallet standard, and is solvable with properly configured peer dependencies

PR #328, opened 2024-06-17
Add wallet sorting props to MUI wallet selector

blakezimmerman:
- Added `sortDefaultWallets` and `sortMoreWallets` props to `WalletConnector`.
- Exposed `WalletConnectorProps` interface.

This will be used to place Petra at the top of the wallets list in Aptos Names and Explorer.

### DIFF
--- a/.changeset/great-suns-yell.md
+++ b/.changeset/great-suns-yell.md
@@ -2,4 +2,4 @@
 "@aptos-labs/wallet-adapter-mui-design": minor
 ---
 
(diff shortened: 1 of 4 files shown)


0xmaayan:
Do dapps manually config the order of the wallets?

blakezimmerman:
> Do dapps manually config the order of the wallets?

They can with these props, but it's optional.

0xmaayan (approved):

PR #330, opened 2024-06-20
Move wallet utils to core package

blakezimmerman:
This PR moves some utilities from the `wallet-adapter-react` package to `wallet-adapter-core` since they're just pure function that don't have any direct relation to React. This is not a breaking change for the `wallet-adapter-react` package since the package already re-exports everything from `wallet-adapter-core`.


packages/wallet-adapter-core/src/utils/helpers.ts
```diff
@@ -105,3 +107,48 @@ export const fetchDevnetChainId = async (): Promise<number> => {
...
+ */
+export function isInstallRequired(wallet: AnyAptosWallet) {
+  const isWalletReady = isInstalledOrLoadable(wallet);
+  const isMobile = !isWalletReady && isRedirectable();
+
+  return !isMobile && !isWalletReady;
+}
+
+/** Truncates the provided wallet address at the middle with an ellipsis. */
+export function truncateAddress(address: string | undefined) {
```
0xmaayan, line 151: lets also update [antd](https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-ant-design/src/WalletSelector.tsx#L12) and [mui](https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-mui-design/src/WalletButton.tsx#L52C17-L52C32) to import from here

0xmaayan (approved):

packages/wallet-adapter-core/src/utils/helpers.ts
blakezimmerman: I already updated MUI to use it last week. I just pushed a commit to make the same change in Antd. The Antd wallet adapter still needs a lot more refactoring to be at par with the shadcn and MUI wallet selectors, but I'll do that in my next PR.

PR #334, opened 2024-06-24
Add AptosConnect Id

kent-white:
Summary

The aptosconnect id wasn't being passed along

 Test Plan

next example still works as expected

### DIFF
--- a/packages/wallet-adapter-core/src/AIP62StandardWallets/sdkWallets.ts
+++ b/packages/wallet-adapter-core/src/AIP62StandardWallets/sdkWallets.ts
@@ -2,8 +2,9 @@ import { DevTWallet, TWallet } from "@atomrigslab/aptos-wallet-adapter";
 import { AptosConnectWallet } from "@aptos-connect/wallet-adapter-plugin";
 import { AptosStandardWallet } from "./WalletStandard";
 import { Network } from "@aptos-labs/ts-sdk";
+import { DappConfig } from "../WalletCore";
 
-export function getSDKWallets(dappConfig?: { network: Network }) {
+export function getSDKWallets(dappConfig?: DappConfig) {
   const sdkWallets: AptosStandardWallet[] = [];
   // Push production wallet if env is production, otherwise use dev wallet
   if (dappConfig?.network === Network.MAINNET) {
@@ -16,7 +17,7 @@ export function getSDKWallets(dappConfig?: { network: Network }) {
 
   // Need to check window is defined for AptosConnect
   if (typeof window !== "undefined") {
-    sdkWallets.push(new AptosConnectWallet({ network: dappConfig?.network }));
+    sdkWallets.push(new AptosConnectWallet({ network: dappConfig?.network, dappId: dappConfig?.aptosConnectDappId }));
   }
 
   return sdkWallets;
--- a/packages/wallet-adapter-core/src/WalletCore.ts
+++ b/packages/wallet-adapter-core/src/WalletCore.ts
@@ -84,6 +84,11 @@ import { aptosStandardSupportedWalletList } from "./AIP62StandardWallets/registr
 
 export type IAptosWallet = AptosStandardWallet & Wallet;
 
+export interface DappConfig {
+  network: Network;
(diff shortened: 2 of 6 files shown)


hardsetting (approved):

0xmaayan (approved):
Please run `changeset` to include it in the next release.

Also, can we update the react package README file with this update? https://github.com/aptos-labs/aptos-wallet-adapter/tree/main/packages/wallet-adapter-react#available-provider-props

packages/wallet-adapter-core/src/AIP62StandardWallets/sdkWallets.ts
```diff
@@ -2,8 +2,9 @@ import { DevTWallet, TWallet } from "@atomrigslab/aptos-wallet-adapter";
 import { AptosConnectWallet } from "@aptos-connect/wallet-adapter-plugin";
 import { AptosStandardWallet } from "./WalletStandard";
 import { Network } from "@aptos-labs/ts-sdk";
+import { DappConfig } from "../WalletCore";
 
-export function getSDKWallets(dappConfig?: { network: Network }) {
+export function getSDKWallets(dappConfig?: DappConfig) {
```
0xmaayan, line 7: nit
```
const { network, dappId } = dappConfig
```

packages/wallet-adapter-react/src/WalletProvider.tsx
```diff
@@ -27,13 +27,14 @@ import type {
   AvailableWallets,
 } from "@aptos-labs/wallet-adapter-core";
 import { WalletCore } from "@aptos-labs/wallet-adapter-core";
+import { DappConfig } from "@aptos-labs/wallet-adapter-core/dist/WalletCore";
```
0xmaayan: after you exported the type -

```suggestion
import { DappConfig } from "@aptos-labs/wallet-adapter-core";
```

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -83,6 +83,11 @@ import { aptosStandardSupportedWalletList } from "./AIP62StandardWallets/registr
 
 export type IAptosWallet = AptosStandardWallet & Wallet;
 
+export interface DappConfig {
```
0xmaayan, line 87: youd need to explicitly export this type 
https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-core/src/index.ts#L3

so it will get exported from the react package so dapps can use it, we already export everything from the core package https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-react/src/index.tsx#L1

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -83,6 +83,11 @@ import { aptosStandardSupportedWalletList } from "./AIP62StandardWallets/registr
...
+  network: Network;
+  aptosConnectDappId?: string;
```
0xmaayan, line 89: curios, whats happens if dapp doesnt provide a dappID for AptosConnect?

packages/wallet-adapter-react/src/WalletProvider.tsx
0xmaayan: I think youd need to export the DappConfig type from the core package to use it.

0xmaayan (changes requested):
we need to export the new type so 
- dapps can use it
- react package to import it from the core package and not from the dist folder

packages/wallet-adapter-core/src/WalletCore.ts
kent-white: the IC legacy accounts aren't shown.

packages/wallet-adapter-core/src/AIP62StandardWallets/sdkWallets.ts
kent-white: i dont think you can destructure optional nicely like this?

0xmaayan (approved):

PR #335, opened 2024-06-24
Update Ant Design wallet selector to new design

blakezimmerman:
This PR updates the Ant Design wallet selector to the new Aptos Connect focused design that we're using for the shadcn and MUI wallet selectors.

## Before

before

## After

after


packages/wallet-adapter-ant-design/src/WalletSelector.tsx
```diff
@@ -40,107 +73,133 @@ export function WalletSelector({
...
-    if (setModalOpen) {
-      setModalOpen(false);
-    }
-  };
-  const buttonText = account?.ansName
-    ? account?.ansName
-    : truncateAddress(account?.address);
+
+  const buttonText =
+    account?.ansName || truncateAddress(account?.address) || "Unknown";
```
0xmaayan, line 84: interesting... would "unknown" ever happen?

0xmaayan (approved):

packages/wallet-adapter-ant-design/src/WalletSelector.tsx
blakezimmerman: I've never seen it happen. I think this is mostly a lacking in our types. We only show this text if a wallet is connected and if `connected` is true then `account` is defined, but TypeScript doesn't understand that connection. I mostly just put this here to cover our bases since previously button text would otherwise be undefined.

PR #343, opened 2024-06-25
Fix for Adapter event communication

0xmaayan:
Currently, the AdapterProvider initializes a WalletCore instance and sets the Wallet state on the first load. Then, the WalletCore constructor executes wallet detection strategies that fire events for the AdapterProvider to listen to. This causes some potential race condition issues, especially when detecting AIP-62 standard wallets, as the events fired from WalletCore might not have registered as listeners in the AdapterProvider yet, and the events are being fired into the void, resulting in wallets not appearing on the Selector Modal.

This PR fixes it by initializing WalletCore in a useEffect hook only once after the initial render and then setting it to a state to have access to WalletCore outside of the useEffect context and to trigger a re-render which ensures the event listeners are reattached and listen to any events coming from WalletCore.

### DIFF
--- a/packages/wallet-adapter-core/src/WalletCore.ts
+++ b/packages/wallet-adapter-core/src/WalletCore.ts
@@ -102,7 +102,7 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
   private _optInWallets: ReadonlyArray<AvailableWallets> = [];
 
   // Private array to hold compatible AIP-62 standard wallets
-  private _standard_wallets: ReadonlyArray<AptosStandardWallet> = [];
+  private _standard_wallets: Array<AptosStandardWallet> = [];
 
   // Private array to hold all wallets (legacy wallet adapter plugins AND compatible AIP-62 standard wallets)
   // while providing support for legacy and new wallet standard
@@ -156,12 +156,16 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
     this._optInWallets = optInWallets;
     this._dappConfig = dappConfig;
     this._sdkWallets = getSDKWallets(this._dappConfig);
+    // Strategy to detect AIP-62 extension standard compatible wallets
+    this.fetchExtensionAIP62AptosWallets();
+    // Strategy to detect AIP-62 SDK standard compatible wallets.
+    // We separate the extension and sdk detection process so we dont refetch sdk wallets anytime a new
+    // extension wallet becomes available
+    this.fetchSDKAIP62AptosWallets();
     // Strategy to detect legacy wallet adapter v1 wallet plugins
     this.scopePollingDetectionStrategy();
-    // Strategy to detect AIP-62 standard compatible wallets (extension + SDK wallets)
-    this.fetchAptosWallets();
     // Append AIP-62 compatible wallets that are not detected on the user machine
-    this.appendNotDetectedStandardSupportedWallets(this._standard_wallets);
+    this.appendNotDetectedStandardSupportedWallets();
   }
 
   private scopePollingDetectionStrategy() {
@@ -188,9 +192,9 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
     });
   }
 
-  private fetchAptosWallets() {
+  private fetchExtensionAIP62AptosWallets() {
     let { aptosWallets, on } = getAptosWallets();
-    this.setWallets(aptosWallets);
+    this.setExtensionAIP62Wallets(aptosWallets);
 
     if (typeof window === "undefined") return;
     // Adds an event listener for new wallets that get registered after the dapp has been loaded,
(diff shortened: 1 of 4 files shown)


packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -239,44 +244,41 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
-   * A function that excludes a wallet the dapp doesnt want to include
+   * A function that excludes an AIP-62 compatible wallet the dapp doesnt want to include
    *
    * @param walletName
    * @returns
    */
   excludeWallet(walletName: string): boolean {
     // for now, we always include AptosConnect
+    // TODO use isAptosConnectWallet
     if (walletName === "Google (AptosConnect)") return false;
```
blakezimmerman: We should probably do something about this `if` branch soon. The current wallet name for Aptos Connect is `Continue with Google`.

packages/wallet-adapter-react/src/WalletProvider.tsx
```diff
@@ -74,18 +75,32 @@ export const AptosWalletAdapterProvider: FC<AptosWalletProviderProps> = ({
...
+      optInWallets ?? [],
+      dappConfig
+    );
+    walletCoreRef.current = walletCore;
+  }, []);
+
+  // Update initial Wallets state once WalletCore has been initialized
+  useEffect(() => {
+    setWallets(walletCoreRef?.current?.wallets ?? []);
+  }, [walletCoreRef?.current]);
```
blakezimmerman: Ref updates don't trigger re-renders, so using it as a `useEffect` dependency probably doesn't behave the way you expect it to. I would recommend storing `walletCore` with `useState` so that this effect is reliably triggered.

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -239,44 +244,41 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
-          aptosStandardWallets.push(wallet);
-          this.standardizeStandardWalletToPluginWalletType(wallet);
-        }
-      }
+  private setExtensionAIP62Wallets(extensionwWallets: readonly AptosWallet[]) {
+    // Twallet SDK fires a register event so the adapter assumes it is an extension wallet
+    // so filter out t wallet, remove it when twallet fixes it
+    const wallets = extensionwWallets.filter(
+      (wallet) => wallet.name !== "Dev T wallet" && wallet.name !== "T wallet"
     );
```
hardsetting, line 271: I think the distinction between "sdk wallet" and "extension wallet" is not necessarily the most useful, and it shows here.
The two should be equivalent as far as the wallet adapter goes and the event-based registration enables this abstraction.

I'll explain myself:

If a wallet is available as chrome extension, it has a `contentscript` it can use to automatically register itself as available wallet.

In case of non-extension wallets, the expectation (at least the way I was envisioning it) is for the dapp developer to register manually.
In AptosConnect I defined a `registerAptosConnect` [function](https://github.com/aptos-labs/aptos-connect/blob/495309e35df5cb6d5493a93ca43cb552a544a596/packages/wallet-adapter-plugin/src/index.ts#L11-L14) the dapp developer can just call in the gobal scope at any point in time, and that's all they need to do.

Of course AptosConnect is a special case, and we want to have it available by default.
So maybe a better distinction would be "detected wallets" vs "default wallets" or "builtin wallets"?
Not sure about "T Wallet" though, you probably have more context on it

packages/wallet-adapter-core/src/WalletCore.ts
hardsetting: Also, we should allow developers to opt-out if they want?
Why are we preventing this

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -289,6 +291,29 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
+    if (this.excludeWallet(wallet.name)) {
+      return;
+    }
+    const isValid = isWalletWithRequiredFeatureSet(wallet);
+    if (isValid) {
+      wallet.readyState = WalletReadyState.Installed;
+      this.standardizeStandardWalletToPluginWalletType(wallet);
+      this._standard_wallets.push(wallet);
+    }
+  }
```
hardsetting: We talked about this before offline, I think we should adapt the old standard into the new one, and not the other way around.

Let's switch it up at some point later

packages/wallet-adapter-react/src/WalletProvider.tsx
hardsetting: Yes agreed. What are you trying to achieve with switching to `useRef` @0xmaayan ?

hardsetting (changes requested):
See @blakezimmerman's comment on `useRef`.

Other than that, left some general comment that can be addressed in later PRs.

packages/wallet-adapter-core/src/WalletCore.ts
0xmaayan: https://aptos-org.slack.com/archives/C051U2U70GK/p1717782986843239?thread_ts=1717528755.480629&cid=C051U2U70GK

packages/wallet-adapter-react/src/WalletProvider.tsx
0xmaayan: from the PR description
```
.....and then setting it to a global ref to have access to WalletCore outside of the useEffect context
```

packages/wallet-adapter-core/src/WalletCore.ts
0xmaayan: The issue with Twallet here, is that they fire a register event which the dapp listens to, but since the adapter already includes the wallet package, the adapter "thinks" there are 2 Twallets - ones coming from the npm dependency and ones coming from the register event.

A npm/sdk wallet, that is already installed and included in WalletCore doesnt need to fire any event to tell the browser "hi i am here" as it is already installed as a dependency.

packages/wallet-adapter-core/src/WalletCore.ts
0xmaayan: This should be a separate work, in a different PR and them released with a major upgrade

packages/wallet-adapter-core/src/WalletCore.ts
0xmaayan: > We should probably do something about this `if` branch soon. The current wallet name for Aptos Connect is `Continue with Google`.

Will change it to `Continue with Google` and then we can use `isAptosConnectWallet` in a next PR. This change requires some type changes also so prefer doing it in a later PR

0xmaayan: this was fixed in https://github.com/aptos-labs/aptos-wallet-adapter/pull/344

packages/wallet-adapter-react/src/WalletProvider.tsx
```diff
@@ -74,18 +67,31 @@ export const AptosWalletAdapterProvider: FC<AptosWalletProviderProps> = ({
   // https://github.com/aptos-labs/aptos-wallet-adapter/issues/94
   const [isLoading, setIsLoading] = useState<boolean>(true);
 
-  const walletCore = useMemo(
-    () => new WalletCore(plugins ?? [], optInWallets ?? [], dappConfig),
-    []
-  );
+  const [walletCore, SetWalletCore] = useState<WalletCore>();
```
hardsetting: typo: setWalletCore

hardsetting (approved):

PR #347, opened 2024-07-01
Add Aptos Connect education screens to wallet selectors

blakezimmerman:
This PR adds an `AboutAptosConnect` headless component to `wallet-adapter-react` and uses it to implement the Aptos Connect education screens in all of our wallet selectors.

This PR also adds documentation for all of the headless components and utilities that have been added recently.

https://github.com/aptos-labs/aptos-wallet-adapter/assets/25761639/0bbb0d17-6632-4eef-8542-190db3ad053b


.changeset/dull-zoos-sell.md
```diff
@@ -0,0 +1,7 @@
+---
+"@aptos-labs/wallet-adapter-ant-design": minor
```
0xmaayan, line 2: if there is no way for end users to disable those screens, we might want to go with a major bump

apps/nextjs-example/src/components/WalletSelector.tsx
```diff
@@ -35,7 +44,23 @@ import {
 } from "./ui/dropdown-menu";
 import { useToast } from "./ui/use-toast";
 
-export function WalletSelector() {
+export interface WalletSelectorProps {
```
0xmaayan: thoughts on this lives in the react package and each selector package imports it? looks like we keep reimplementing it on any selector ui package

.changeset/dull-zoos-sell.md
blakezimmerman: Currently, Aptos Connect can be disabled by not passing it to the `optInWallets` prop on the `AptosWalletProvider` component — this would also hide the education screens. I think we generally want projects to always display the education screens when Aptos Connect is available, but let me know what you think.

apps/nextjs-example/src/components/WalletSelector.tsx
blakezimmerman: It would be pretty difficult to make a generic version of this one component. The Radix, Ant, and MUI implementation each use different styling solutions. Radix uses Tailwind, Ant uses vanilla CSS, and MUI uses Emotion (via the sx prop). To move this into `wallet-adapter-react`, this component would have to accept a `className` prop for every single element in the selector and it's still not clear to me how I would handle the `asChild` pattern I'm using to render the headless components as Radix, Ant, and MUI specific components. I think using a composition model with these headless components is a better approach than trying to write a component with dozens of props that tries to satisfy each implementation.

.changeset/dull-zoos-sell.md
0xmaayan: > Currently, Aptos Connect can be disabled by not passing it to the `optInWallets` prop on the `AptosWalletProvider` component — this would also hide the education screens. I think we generally want projects to always display the education screens when Aptos Connect is available, but let me know what you think.

actually, for now, [we always include AptosConnect](https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-core/src/WalletCore.ts#L285).

I agree we probably want projects to always display it, but it should be their decision eventually, and they should be able to simply do it by upgrading the react package

apps/nextjs-example/src/components/WalletSelector.tsx
0xmaayan: oh I was only talking about the WalletSelectorProps interface

.changeset/dull-zoos-sell.md
blakezimmerman: Okay, I'll make this a major version bump then.

blakezimmerman: I updated the changeset

apps/nextjs-example/src/components/WalletSelector.tsx
blakezimmerman: Oh I see. I'll think about how we should share this. I could just drop this interface into the shared package, but it might a bit strange to export an interface that isn't tied to any specific implementation in the package. Maybe I'll add another utility to the shared package that consolidates all the below logic into a single function:

```ts
const {
  /** Wallets that use social login to create an account on the blockchain */
  aptosConnectWallets,
  /** Wallets that use traditional wallet extensions */
  otherWallets,
} = getAptosConnectWallets(wallets);

const {
  /** Wallets that are currently installed or loadable. */
  defaultWallets,
  /** Wallets that are NOT currently installed or loadable. */
  moreWallets,
} = partitionWallets(otherWallets);

if (sortDefaultWallets) defaultWallets.sort(sortDefaultWallets);
if (sortMoreWallets) moreWallets.sort(sortMoreWallets);
```

packages/wallet-adapter-react/docs/BYO-wallet-selector.md
```diff
@@ -0,0 +1,205 @@
+# Build Your Own Wallet Selector
```
0xmaayan, line 1: we probably want to include it or at least mention it on aptos.dev cc @jmintuitive

0xmaayan (approved):

apps/nextjs-example/src/components/WalletSelector.tsx
blakezimmerman: I just added a `groupAndSortWallets` function to `wallet-adapter-core` that encapsulates the grouping and sorting logic while providing an interface that includes all of those sorting options that were previously copy and pasted across the wallet selector implementations.

0xmaayan (approved):
LGTM

PR #350, opened 2024-07-02
Add docs to Wallet Adapter AIP-62 plugin template

jmintuitive:
Added explanations of how to rewrite the AIP-62 plugin template in order to implement your own version. 

Open questions I still want to include (@0xmaayan maybe you know the answers?)
1. How can they test whether their code works? (Should they try their template within this repo? If so, what should they change?)
2. What should they do once they've rewritten all this code to "deploy" it so dapps can use it? (Do they need to update a certain folder with their updated code?)
3. Are there any examples of 3rd party wallet's implementations here that I can link to as further examples?

(I'd also love in-depth review of the comments to make sure they're accurate - especially for what the implementations of each feature should actually do).


apps/nextjs-example/src/utils/standardWallet.ts
```diff
@@ -32,51 +32,157 @@ import {
...
+  /**
+   * REVISION - Set the subset of Aptos chains your wallet supports.
+   * APTOS_CHAINS = ["aptos:devnet", "aptos:testnet", "aptos:localnet", "aptos:mainnet"]
+   * Many wallets only support "aptos:mainnet" and "aptos:testnet".
```
0xmaayan: the recommendation is to at least support mainnet, testnet and devnet

apps/nextjs-example/src/utils/standardWallet.ts
```diff
@@ -221,16 +395,33 @@ export class MyWallet implements AptosWallet {
...
+   * @returns when the logic is resolved.
+   */
   onNetworkChange: AptosOnNetworkChangeMethod = async (): Promise<void> => {
+    // THIS LOGIC SHOULD BE REPLACED. IT IS FOR EXAMPLE PURPOSES ONLY.    
     return Promise.resolve();
   };
 }
 
-// a wallet call to register itself (this happens on the wallet side)
+// This section should NOT be revised (although it may make sense to move where it gets called).
```
0xmaayan: so this section is (only) for chrome extension wallets (the wallets we need to install from the chrome store). If the wallet is a SDK wallet (npm package) there is no need to fire a registerWallet event.

We use it here just so the demo dapp can be aware of this wallet example

0xmaayan:
> * How can they test whether their code works? (Should they try their template within this repo? If so, what should they change?)
> * What should they do once they've rewritten all this code to "deploy" it so dapps can use it? (Do they need to update a certain folder with their updated code?)
> * Are there any examples of 3rd party wallet's implementations here that I can link to as further examples?

The very naive approach to test their code is to fork the repo, implement the class, import the [standardWallet.ts](https://github.com/aptos-labs/aptos-wallet-adapter/pull/350/files#diff-2765f3eb274fea4a1bf14a2beee67d79ffe4844b9e8876753c7b876b9825eeec) file and test it with the local demo dapp.

The better approach would be to actually develop their wallet, and then based on whether it is a chrome extension or sdk wallet, to test it with a live dapp (like https://aptos-labs.github.io/aptos-wallet-adapter/ )

As for a real wallet code example, @hardsetting @kent-white have we open source Petra?

apps/nextjs-example/src/utils/standardWallet.ts
jmintuitive: Is there a separate example of an SDK wallet template? (How does the npm package version register itself?)

(Maybe we can chat about this later in a bit more depth?)

0xmaayan: The SDK/npm package adds itself to the adapter code https://aptos.dev/en/build/sdks/wallet-adapter/wallets#is-your-wallet-a-sdk-npm-package-wallet

0xmaayan (approved):

apps/nextjs-example/src/utils/standardWallet.ts
```diff
@@ -32,40 +32,93 @@ import {
 } from "@aptos-labs/wallet-standard";
 
 /**
- * A class to create a mock wallet for demonstration a wallet
- * implementation compatible with Aptos AIP-62 Wallet Standard
+ * This class is a template you can modify to implement an AIP-62 Wallet.
```
0xmaayan: this is a copy of https://github.com/aptos-labs/wallet-standard/blob/main/example/wallet.ts - instead of keep maintaining both docs, can we link to the standard repo here instead?

jmintuitive: Yep! Added a link to that and updated the docs so they are just documenting the code, not recommending revisions (which is what the template is for).

0xmaayan (approved):

PR #356, opened 2024-07-10
Use example plugin in next.js demo app

jmintuitive:
For people writing their own wallet-adapter-plugin, it's helpful to see in practice how to register their code.

This PR updates the example app to show that `registerWallet` logic and allow users to play with the "Example Wallet". 

It also fixes the example wallet + next.js app implementation to avoid the `fundWallet` bug.


apps/nextjs-example/src/app/page.tsx
```diff
@@ -34,6 +34,19 @@ import {
...
 import Image from "next/image";
 
+// Imports for registering a browser extension wallet plugin on page load
+import { MyWallet } from "@/utils/standardWallet";
+import { registerWallet } from "@aptos-labs/wallet-standard";
+
+// Example of how to register a browser extension wallet plugin.
+// Browser extension wallets should call registerWallet once on page load.
+// When you click "Connect Wallet", you should see "Example Wallet" 
+(function () {
```
0xmaayan, line 44: so that adds the "Example Wallet" to the UI as a wallet option, have we tested it on the demo dapp ui?

jmintuitive: The Example Wallet works as well as Petra does in the UI :) 

There is a bug I found when testing that the Multiagent flow fails on Submit in particular. I'm not entirely sure why - but it seems that there is an INVALID_AUTHENTICATOR error. 

Screenshot: 
image

For now I updated the error handling so it doesn't crash the app (instead it just shows the normal "failed to submitTransaction" toast). I spent about an hour trying to debug and didn't want to invest more time unless you think it's worth me focusing on it :)

0xmaayan: thats fine, no need to spend time investigating bugs but focus on writing =) feel free to open a bug issue on this repo

0xmaayan (approved):

apps/nextjs-example/src/app/page.tsx
jmintuitive: Done :) - https://github.com/aptos-labs/aptos-wallet-adapter/issues/364

PR #357, opened 2024-07-10
Update the docs in wallet-adapter-core for AIP62

jmintuitive:
Adds additional information for what should go into each field in the AIP62StandardWallet folder.


packages/wallet-adapter-core/src/AIP62StandardWallets/sdkWallets.ts
```diff
@@ -21,5 +21,8 @@ export function getSDKWallets(dappConfig?: DappConfig) {
     sdkWallets.push(new DevTWallet() as any);
   }
 
+  // Add new SDK wallet plugins (ones that should be installed as packages) here:
```
0xmaayan, line 24: not completely sure this is needed

0xmaayan (approved):

packages/wallet-adapter-core/src/AIP62StandardWallets/sdkWallets.ts
jmintuitive: I think it's optional, but I liked the idea of flagging where people should modify the code as it just helps re-assure people they're modifying the right part of the file :)

PR #365, opened 2024-07-17
Add examples of how to use the React component

jmintuitive:
These examples were written for the aptos.dev content, but are useful to have in the codebase so they can easily be imported into the example dapp for users to play with the individual features without the complexity of the custom UI components.


rangel0828:
Acceder

PR #366, opened 2024-07-17
Add the proper type for the dapp config in wallet core

kent-white:
This was working from wallet-adapter-react because it gets casted into dappconfig below, but when using wallet adapter core directly you would run into a type error.


hardsetting (approved):

.changeset/violet-pigs-impress.md
hardsetting, line 1: nice name

PR #368, opened 2024-07-19
support gate wallet

vdgwerdfgCell:


0xmaayan:
Looks like you implemented the legacy wallet standard - highly recommend to implement and be compatible with the new wallet standard so the wallet will work with any dapp and provide a much better user experience

Can follow the instructions here: https://aptos.dev/en/build/sdks/wallet-adapter/browser-extension-wallets

vdgwerdfgCell:
> Looks like you implemented the legacy wallet standard - highly recommend to implement and be compatible with the new wallet standard so the wallet will work with any dapp and provide a much better user experience
> 
> Can follow the instructions here: https://aptos.dev/en/build/sdks/wallet-adapter/browser-extension-wallets

vdgwerdfgCell:
Can we support legacy wallet standard first, we will support the new wallet standard next time. Thanks sir please

PR #373, opened 2024-07-23
[Hot Fix] Pontem wallet version 2.6.7 does not include a URL prop

0xmaayan:
Pontem released a new Wallet version and it does not include a `URL` prop - that breaks some of the adapter internal checks.

Additionally, they changes their `name` prop from `Pontem` to `Pontem Wallet` and that also breaks some of the adapter checks on how to decide what wallets to show on the Wallet Selector so it wont have duplications

### DIFF
--- a/packages/wallet-adapter-core/src/WalletCore.ts
+++ b/packages/wallet-adapter-core/src/WalletCore.ts
@@ -180,6 +180,15 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
       if (typeof window !== "undefined") {
         scopePollingDetectionStrategy(() => {
           const providerName = wallet.providerName || wallet.name.toLowerCase();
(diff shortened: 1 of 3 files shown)


packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -180,6 +180,15 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
       if (typeof window !== "undefined") {
         scopePollingDetectionStrategy(() => {
           const providerName = wallet.providerName || wallet.name.toLowerCase();
+          // Hot fix to manage Pontem wallet to not show duplications if user has the
+          // new standard version installed. Pontem uses "Pontem" wallet name for previous versions and
```
bv-aptos, line 184: nit: comment says "Pontem" but the check is for "Pontem Wallet"

bv-aptos (approved):

PR #380, opened 2024-07-29
Bypass verification for Keyless wallet accounts

heliuchuan:
To be removed when verification in the SDK is done.  For now, instead of erroring, just return true.


0xmaayan (approved):
please run changeset https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/CONTRIBUTING.md#changesets

PR #384, opened 2024-07-31
Fix turbo config following a new turbo version

0xmaayan:
Turbo upgraded to a new version and made changes to its config file, we now use `tasks` instead of `pipeline`. 
https://github.com/vercel/turbo/pull/8157

This upgrade fails [PR builds as CI](https://github.com/aptos-labs/aptos-wallet-adapter/actions/runs/10181964067/job/28165000240) uses latest turbo version while locally we still use the old version.

This PR upgrades turbo and adjusts the config file.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -17,19 +17,19 @@ importers:
         version: link:packages/eslint-config-adapter
       '@changesets/cli':
         specifier: ^2.22.0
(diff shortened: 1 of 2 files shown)


kent-white (approved):

hardsetting:
I think we should also pin the major version of turbo, so there's never a mismatch in the future :)

PR #398, opened 2024-08-14
Have the option to exclude AptosConnect from the wallet selector modal

0xmaayan:
AptosConnect is currently included by default in the wallet selector modal - this causes issues for dapps that run on platforms AptosConnect is yet supported on (TG, file upload service, etc)

This change provides the option for devs to not opt-in to AptosConnect

### DIFF
--- a/apps/nextjs-example/src/components/ui/use-toast.ts
+++ b/apps/nextjs-example/src/components/ui/use-toast.ts
@@ -83,7 +83,7 @@ export const reducer = (state: State, action: Action): State => {
       return {
         ...state,
         toasts: state.toasts.map((t) =>
-          t.id === action.toast.id ? { ...t, ...action.toast } : t,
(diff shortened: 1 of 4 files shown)


apps/nextjs-example/src/components/ui/use-toast.ts
```diff
@@ -155,7 +155,7 @@ function toast({ ...props }: Toast) {
       ...props,
       id,
       open: true,
-      onOpenChange: (open) => {
+      onOpenChange: (open: any) => {
```
blakezimmerman, line 158: Why this `any` annotation?

0xmaayan: it started to complain on not having a type and I couldn't build, so just quickly added any

gregnazario (approved):

PR #418, opened 2024-09-13
[core] Bumped `@aptos-connect/wallet-adapter-plugin` to 2.0.2

hardsetting:
... and exposed `preferredWalletName` in the adapter config

https://github.com/user-attachments/assets/86cc05ae-57ed-46cf-967b-6dfa81b1a6e8

### DIFF
--- a/packages/wallet-adapter-core/src/WalletCore.ts
+++ b/packages/wallet-adapter-core/src/WalletCore.ts
@@ -82,12 +82,16 @@ import {
 import { GA4 } from "./ga";
 import { WALLET_ADAPTER_CORE_VERSION } from "./version";
 import { aptosStandardSupportedWalletList } from "./AIP62StandardWallets/registry";
+import type { AptosConnectWalletConfig } from "@aptos-connect/wallet-adapter-plugin";
 
 export type IAptosWallet = AptosStandardWallet & Wallet;
 
 export interface DappConfig {
   network: Network;
+  /** @deprecated */
   aptosConnectDappId?: string;
+  // Config used to initialize the AptosConnect wallet provider
+  aptosConnect?: Omit<AptosConnectWalletConfig, 'network'>,
   mizuwallet?: {
     manifestURL: string;
     appId?: string;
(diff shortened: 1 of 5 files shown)


packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -87,7 +87,12 @@ export type IAptosWallet = AptosStandardWallet & Wallet;
 
 export interface DappConfig {
   network: Network;
+  /** @deprecated */
   aptosConnectDappId?: string;
+  aptosConnect?: {
+    dappId?: string;
```
0xmaayan: nit: can we add a doc string above `aptosConnect` prop explaining each of the property options?

0xmaayan (approved):

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -82,12 +82,16 @@ import {
 import { GA4 } from "./ga";
 import { WALLET_ADAPTER_CORE_VERSION } from "./version";
 import { aptosStandardSupportedWalletList } from "./AIP62StandardWallets/registry";
+import type { AptosConnectWalletConfig } from "@aptos-connect/wallet-adapter-plugin";
 
 export type IAptosWallet = AptosStandardWallet & Wallet;
 
 export interface DappConfig {
   network: Network;
+  /** @deprecated */
```
0xmaayan, line 91: if this is deprecated, what should we put here? https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/apps/nextjs-example/src/components/WalletProvider.tsx#L36

hardsetting: we could put
"aptosConnect: { dappId }"
instead, but wanted to keep that value there also to make sure it still works even if deprecated

0xmaayan: ohh I see, 
```
export interface AptosConnectWalletConfig extends Omit<ACDappClientConfig, 'defaultNetworkName'> {
  network?: Network;
  preferredWalletName?: string;
}
```

and `ACDappClientConfig` has a `dappId` prop

PR #425, opened 2024-09-17
Add support for custom chains in changeNetwork function

arjanjohan:
### Current Situation:
The changeNetwork function does not support custom networks because the chainId cannot be inferred from the network name.

### Fix:
This PR introduces an optional customChainId parameter to the changeNetwork function in `WalletCore.ts`. The parameter is only used when switching to a custom chain.

Additionally, I have updated both `wallet-adapter-react` and `wallet-adapter-vue` to pass this optional parameter when calling the changeNetwork function.

### Tests:
I updated `apps/nextjs-example`, and it now supports switching between both custom and predefined networks. However, I did not update `apps/nuxt-example` as I have limited experience with Nuxt, and network switching in that example already appears to be broken.


packages/wallet-adapter-react/src/WalletProvider.tsx
```diff
@@ -186,18 +186,21 @@ export const AptosWalletAdapterProvider: FC<AptosWalletProviderProps> = ({
...
 
-  const changeNetwork = async (network: Network) => {
+  const changeNetwork = async (network: Network, chainId?: number) => {
     if (!walletCore) {
       throw new Error("WalletCore is not initialized");
     }
+    
     try {
-      return await walletCore?.changeNetwork(network);
+      return chainId !== undefined
```
0xmaayan: we dont really need this conditional logic as we accept an undefined customChainId and handle it in the wallet-core logic https://github.com/aptos-labs/aptos-wallet-adapter/pull/425/files#diff-8a76cc616d7961765eb7ff494b4e86bc0e56cce6d0273caaa57af714ac95757eR1178
```suggestion
      return await walletCore.changeNetwork(network, chainId)
```

packages/wallet-adapter-react/src/useWallet.tsx
```diff
@@ -39,7 +39,7 @@ export interface WalletContextState {
   ): Promise<PendingTransactionResponse>;
   signMessage(message: SignMessagePayload): Promise<SignMessageResponse>;
   signMessageAndVerify(message: SignMessagePayload): Promise<boolean>;
-  changeNetwork(network: Network): Promise<AptosChangeNetworkOutput>;
+  changeNetwork(network: Network, chainId?: number): Promise<AptosChangeNetworkOutput>;
```
0xmaayan: Lets let the user pass in the chainId as a string, and have the adapter do the parsing before sending it to the wallet

0xmaayan (commented):
Thanks for the contribution! Overall LGTM, left some comments

packages/wallet-adapter-react/src/useWallet.tsx
arjanjohan: I know in the example used in this repo, the chainId is provided in an input field and using string would make sense. However, I believe this to be an exceptional case. I imagina all dApps that will utilize this feature will handle the logic in the backend, and the chainId is not provided by the user but rather from some backend storage where it's stored as number.

Let me know what you think.

packages/wallet-adapter-vue/src/composables/useWallet.ts
```diff
@@ -197,9 +197,12 @@ export function useWallet(
...
   };
 
-  const changeNetwork = async (network: Network) => {
+  const changeNetwork = async (network: Network, chainId?: number) => {
     try {
-      return await walletCoreInstance.changeNetwork(network);
+      // Call changeNetwork with chainId if provided, otherwise call without it
+      return chainId !== undefined 
+        ? await walletCoreInstance.changeNetwork(network, chainId) 
+        : await walletCoreInstance.changeNetwork(network);
```
0xmaayan:
```suggestion
      await walletCoreInstance.changeNetwork(network, chainId) 
```

packages/wallet-adapter-react/src/useWallet.tsx
0xmaayan: I see. I'd argue that they will be retrieving it from an API query, where it will mostly be in a string format. Instead of requiring the [dapp to convert it to a number](https://github.com/aptos-labs/aptos-wallet-adapter/pull/425/files#diff-010ca2e223c5ddff67f01665c9504e1c74c1b0c36861e9b6441e6a13f11b0675R299-R300), the [adapter can handle](https://github.com/aptos-labs/aptos-wallet-adapter/pull/425/files#diff-8a76cc616d7961765eb7ff494b4e86bc0e56cce6d0273caaa57af714ac95757eR1178) that conversion for them.

0xmaayan (commented):
Please run [changeset](https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/CONTRIBUTING.md#changesets) to include this change in the next release

packages/wallet-adapter-react/src/useWallet.tsx
arjanjohan: That makes sense, thanks for clarifying. I will modify it and push the changes.

apps/nextjs-example/src/app/page.tsx
```diff
@@ -162,6 +163,8 @@ function WalletConnection({
     return true;
   };
 
+  const [customChainId, setCustomChainId] = useState<string | "">("");
```
0xmaayan:
```suggestion
  const [customChainId, setCustomChainId] = useState<string>("");
```

0xmaayan (commented):
Thanks! LGTM.
Only two comments are left to address, and then we can merge.

arjanjohan:
> Thanks! LGTM. Only two comments are left to address, and then we can merge.

I processed the comments and updated the example nextJS app.

0xmaayan (approved):

arjanjohan:
@0xmaayan bump. I noticed there was a merge conflict blocking this PR. Resolved that now and it's good to go.

PR #429, opened 2024-09-25
[nextjs] fix multiAgent  transfer script

WGB5445:


apps/nextjs-example/src/components/transactionFlows/MultiAgent.tsx
```diff
@@ -78,8 +83,16 @@ export function MultiAgent() {
     if (!transactionToSubmit) {
       throw new Error("No Transaction to sign");
     }
+    if (!secondarySignerAccount) {
+      throw new Error("No secondarySignerAccount");
+    }
     try {
-      const authenticator = await signTransaction(transactionToSubmit);
+      const authenticator = aptosClient(
```
0xmaayan: what this change for? the idea of using this demo app is to use the wallet adapter to sign a transaction and not the sdk

apps/nextjs-example/src/components/transactionFlows/MultiAgent.tsx
```diff
@@ -14,6 +17,8 @@ import { useToast } from "../ui/use-toast";
 import { LabelValueGrid } from "../LabelValueGrid";
 
 const APTOS_COIN = "0x1::aptos_coin::AptosCoin";
+const TRANSFER_SCRIPT =
+  "0xa11ceb0b0600000006010002030206040802050a10071a1d0837200000000103010100000204060c060c05030001090003060c05030d6170746f735f6163636f756e740e7472616e736665725f636f696e73000000000000000000000000000000000000000000000000000000000000000101000001050b010b020b03380002";
```
0xmaayan: Could we leave a comment with where/what script this is?

0xmaayan (approved):

apps/nextjs-example/src/components/transactionFlows/MultiAgent.tsx
```diff
@@ -55,9 +60,9 @@ export function MultiAgent() {
       sender: account.address,
       secondarySignerAddresses: [secondarySigner.accountAddress],
       data: {
-        function: "0x1::coin::transfer",
-        typeArguments: [APTOS_COIN],
-        functionArguments: [account.address, 1], // 1 is in Octas
+        bytecode: TRANSFER_SCRIPT,
```
0xmaayan, line 62: whats the reason we use pre compiled script and not an entry function?

WGB5445: Since there are no on-chain contract functions that support multi-agent, we have to implement a workaround to make this functionality possible.  
Without this approach, the current multi-agent feature would not be usable.

apps/nextjs-example/src/components/transactionFlows/MultiAgent.tsx
WGB5445: That’s totally reasonable.

apps/nextjs-example/src/components/transactionFlows/MultiAgent.tsx
WGB5445: Our example has an issue. Based on the current logic, we can't add a second signer because it's not possible to connect to two wallets at the same time. As a result, when composing the transaction, we can't obtain both the `signer` address and the `secondarySignerAddresses` simultaneously.

So it seems quite difficult to successfully demonstrate this example. You can try running it in the current setup—it will likely throw an error.

According to my earlier idea, we could implement this functionality on devnet and use a randomly generated account as the secondary signer. This way, only one account needs to sign, which simplifies the process.

gregnazario (approved):

apps/nextjs-example/src/components/transactionFlows/MultiAgent.tsx
```diff
@@ -55,9 +66,9 @@ export function MultiAgent() {
       sender: account.address,
       secondarySignerAddresses: [secondarySigner.accountAddress],
       data: {
-        function: "0x1::coin::transfer",
-        typeArguments: [APTOS_COIN],
-        functionArguments: [account.address.toString(), 1], // 1 is in Octas
+        bytecode: TRANSFER_SCRIPT,
+        typeArguments: [parseTypeTag(APTOS_COIN)],
+        functionArguments: [AccountAddress.fromString(account.address), new U64(1)],
```
gregnazario: `account.address` is now an `AccountAddress`
```suggestion
        typeArguments: [parseTypeTag(APTOS_COIN)],
        functionArguments: [account.address, new U64(1)],
```

gregnazario (commented):
Please test :)

apps/nextjs-example/src/components/transactionFlows/MultiAgent.tsx
WGB5445: fixed

PR #432, opened 2024-10-04
Handle arguments conversion when submitting a package publish transac…

0xmaayan:
…tion

The Aptos SDK expects the `metadataBytes` and `byteCode` to be `Uint8Array`, but in case the arguments are passed in
as a string, this function converts the string to Uint8Array. 

Now we can use the wallet adapter to submit a package publish transaction like

```
const response = await signAndSubmitTransaction({
      data: {
        function: "0x1::code::publish_package_txn",
        functionArguments: [
         "0x123",
          ["0x456"],
        ],
      },
    });
```

and
```
const response = await signAndSubmitTransaction({
      data: {
        function: "0x1::code::publish_package_txn",
        functionArguments: [
          Hex.fromHexInput( "0x123").toUint8Array(),
          [Hex.fromHexInput("0x456").toUint8Array()],
        ],
      },
    });
```

https://aptos-org.slack.com/archives/C05161WELKZ/p1726562237760779

### DIFF
--- a/.changeset/new-readers-develop.md
+++ b/.changeset/new-readers-develop.md
@@ -1,0 +1,5 @@
+---
+"@aptos-labs/wallet-adapter-core": minor
+---
+
+Handle arguments conversion when submitting a package publish transaction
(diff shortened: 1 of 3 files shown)


gregnazario (approved):
A bit of a workaround, but will definitely get rid of this confusion.

PR #434, opened 2024-10-05
[core] AptosConnect Telegram support

hardsetting:
See https://github.com/aptos-labs/aptos-connect/pull/624

Nothing required from dapp developers other than bumping to the latest version of the Wallet Adapter.

https://github.com/user-attachments/assets/86a459ca-9f61-4e4e-ae72-8dc08b305dee

### DIFF
--- a/apps/nextjs-example/package.json
+++ b/apps/nextjs-example/package.json
@@ -30,6 +30,7 @@
     "@radix-ui/react-slot": "^1.0.2",
     "@radix-ui/react-switch": "^1.0.3",
     "@radix-ui/react-toast": "^1.1.5",
+    "@telegram-apps/sdk": "^2.0.0",
     "@trustwallet/aptos-wallet-adapter": "^0.1.6",
     "antd": "^5.1.2",
     "class-variance-authority": "^0.7.0",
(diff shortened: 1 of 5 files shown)


apps/nextjs-example/package.json
```diff
@@ -30,6 +30,7 @@
     "@radix-ui/react-slot": "^1.0.2",
     "@radix-ui/react-switch": "^1.0.3",
     "@radix-ui/react-toast": "^1.1.5",
+    "@telegram-apps/sdk": "^2.0.0",
```
0xmaayan, line 33: In CAD we use '@telegram-apps/sdk-react' - is it a different package?

https://github.com/aptos-labs/create-aptos-dapp/blob/main/templates/clicker-game-tg-mini-app-template/package.json#L31

0xmaayan (approved):
LGTM, just a question on the package used here versus CAD

apps/nextjs-example/package.json
hardsetting: It's the same underlying package (although I'm using v2 here)

'react-sdk' builds on top of 'sdk'

hardsetting: didn't need full react support

Joj501 (approved):

PR #435, opened 2024-10-14
Improve compositionability a bit in react lib's createHeadlessComponent helper

psd-coder:
This PR improves `createHeadlessComponent` util behavior: if `asChild` is set and `children` is valid react element with empty `children` props (to not override custom children there) then will be used `children` from props.

# Before change
Right now if you use `asChild` for passing props to children you won't get a default children value from the defaults.

For example this won't work and will give empty `<p class="text-primary"></p>`
```
<AptosPrivacyPolicy.Disclaimer asChild>
  <p className="text-primary" />
</AptosPrivacyPolicy.Disclaimer>
```

# After the change

```
<AptosPrivacyPolicy.Disclaimer asChild>
  <p className="text-primary" />
</AptosPrivacyPolicy.Disclaimer>
```
will give you `<p class="text-primary">By continuing, you agree to Aptos Labs'</p>`

and if you specify children in internal element, then

```
<AptosPrivacyPolicy.Disclaimer asChild>
  <p className="text-primary">By continuing, you agree with the policy</p>
</AptosPrivacyPolicy.Disclaimer>
```
will give you `<p className="text-primary">By continuing, you agree with the policy</p>`

### DIFF
--- a/packages/wallet-adapter-react/src/components/utils.tsx
+++ b/packages/wallet-adapter-react/src/components/utils.tsx
@@ -1,5 +1,5 @@
 import { Slot } from "@radix-ui/react-slot";
-import { ReactNode, forwardRef } from "react";
+import { ReactNode, cloneElement, forwardRef, isValidElement } from "react";
 
 export interface HeadlessComponentProps {
   /** A class name for styling the element. */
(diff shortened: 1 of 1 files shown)


0xmaayan (approved):

blakezimmerman (approved):
Nice improvement, thanks!

PR #441, opened 2024-10-30
Refactor the account info output type to be compatible with `AIP-62` standard

0xmaayan:
To support the different Public Key types and for developers to easily be able to pass the account info down to SDK functions, changing the Account Info return type to be compatible with AIP-62

An updated version with this change here: https://aptos-labs.github.io/aptos-wallet-adapter/nextjs-example-testing/

Will be released with a major version update

OLD
```
export type AccountInfo = {
  address: string;
  publicKey: string | string[];
  minKeysRequired?: number;
  ansName?: string | null;
};
```

NEW
```
export type AccountInfoOutput = {
  address: AccountAddress;
  publicKey: PublicKey | PublicKey[]; // PublicKey[] for backward compatibility
  minKeysRequired?: number;
  ansName?: string | null;
};
```

This way, one can do
```
  const signWithSDK = async () => {
    if (!account) return;

    // make sure it is not an array of public keys
    if (Array.isArray(account.publicKey)) return;

    const transaction = await aptosClient(network).transaction.build.simple({
      sender: account.address,
      data: {
        typeArguments: [APTOS_COIN],
        functionArguments: [account.address.toString(), 1], // 1 is in Octas
      },
    });
    const response = await aptosClient(network).transaction.simulate.simple({
      signerPublicKey: account.publicKey,
      transaction,
    });
  };
```

### DIFF
--- a/apps/nextjs-example/src/app/page.tsx
+++ b/apps/nextjs-example/src/app/page.tsx
@@ -25,6 +25,7 @@ import { WalletSelector as AntdWalletSelector } from "@aptos-labs/wallet-adapter
 import { WalletConnector as MuiWalletSelector } from "@aptos-labs/wallet-adapter-mui-design";
 import {
   AccountInfo,
+  AccountInfoOutput,
   AptosChangeNetworkOutput,
   NetworkInfo,
   WalletInfo,
@@ -48,7 +49,9 @@ import { registerWallet } from "@aptos-labs/wallet-standard";
   registerWallet(myWallet);
 })();
 
-const isTelegramMiniApp = typeof window !== 'undefined' && (window as any).TelegramWebviewProxy !== undefined;
+const isTelegramMiniApp =
+  typeof window !== "undefined" &&
+  (window as any).TelegramWebviewProxy !== undefined;
 if (isTelegramMiniApp) {
   initTelegram();
 }
@@ -147,7 +150,7 @@ function WalletSelection() {
 }
 
 interface WalletConnectionProps {
-  account: AccountInfo | null;
+  account: AccountInfoOutput | null;
   network: NetworkInfo | null;
   wallet: WalletInfo | null;
   changeNetwork: (network: Network) => Promise<AptosChangeNetworkOutput>;
@@ -225,7 +228,7 @@ function WalletConnection({
                 label: "Address",
                 value: (
                   <DisplayValue
-                    value={account?.address ?? "Not Present"}
+                    value={account?.address.toString() ?? "Not Present"}
                     isCorrect={!!account?.address}
                   />
                 ),
--- a/packages/wallet-adapter-core/src/LegacyWalletPlugins/types.ts
+++ b/packages/wallet-adapter-core/src/LegacyWalletPlugins/types.ts
@@ -8,7 +8,8 @@ import {
   InputGenerateTransactionPayloadData,
   AnyRawTransaction,
   Signature,
-  AccountAuthenticator,
+  AccountAddress,
+  PublicKey,
 } from "@aptos-labs/ts-sdk";
 import { WalletReadyState } from "../constants";
 import {
@@ -20,7 +21,10 @@ import {
   AptosChangeNetworkMethod,
   AptosSignAndSubmitTransactionInput,
 } from "@aptos-labs/wallet-standard";
-import { AptosStandardSupportedWallet } from "../AIP62StandardWallets/types";
+import {
+  AptosStandardSupportedWallet,
+  StandardAccountInfoInput,
+} from "../AIP62StandardWallets/types";
 
 export { TxnBuilderTypes, Types } from "aptos";
 export type {
@@ -55,6 +59,7 @@ export type WalletInfo = {
   url: string;
 };
 
+// Input type to handle legacy wallets that are not AIP-62 compatible
 export type AccountInfo = {
   address: string;
   publicKey: string | string[];
@@ -74,7 +79,7 @@ export declare interface WalletCoreEvents {
   readyStateChange(wallet: Wallet): void;
   standardWalletsAdded(wallets: Wallet | AptosStandardSupportedWallet): void;
   networkChange(network: NetworkInfo | null): void;
-  accountChange(account: AccountInfo | null): void;
+  accountChange(account: AccountInfo | StandardAccountInfoInput | null): void;
 }
 
 export interface SignMessagePayload {
--- a/packages/wallet-adapter-core/src/WalletCore.ts
+++ b/packages/wallet-adapter-core/src/WalletCore.ts
@@ -8,14 +8,13 @@ import {
   Ed25519PublicKey,
   InputGenerateTransactionOptions,
   Ed25519Signature,
-  AptosConfig,
   InputSubmitTransactionData,
   PendingTransactionResponse,
   Aptos,
   generateRawTransaction,
   SimpleTransaction,
   NetworkToChainId,
-  Hex,
+  PublicKey,
 } from "@aptos-labs/ts-sdk";
 import EventEmitter from "eventemitter3";
 import {
@@ -60,7 +59,6 @@ import {
   WalletName,
   WalletCoreV1,
   CompatibleTransactionOptions,
-  convertNetwork,
   convertPayloadInputV1ToV2,
   generateTransactionPayloadFromV1Input,
 } from "./LegacyWalletPlugins";
@@ -80,6 +78,8 @@ import {
   WalletStandardCore,
   AptosStandardSupportedWallet,
   AvailableWallets,
+  StandardAccountInfoInput,
+  AccountInfoOutput,
 } from "./AIP62StandardWallets";
 import { GA4 } from "./ga";
 import { WALLET_ADAPTER_CORE_VERSION } from "./version";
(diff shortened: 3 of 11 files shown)


packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -783,6 +759,7 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
       } else {
         account = await this.walletCoreV1.connect(selectedWallet);
       }
+      console.log("connectWallet account", account);
```
GhostWalker562: Probably remove?

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -687,9 +642,30 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
+      if (!this._account) return null;
+      return {
+        address:
+          typeof this._account.address === "string"
+            ? AccountAddress.from(this._account.address)
+            : this._account.address,
+        publicKey:
+          // for backward compatibility, if the account public key is of type `string` convert it to Ed25519PublicKey
+          typeof this._account.publicKey === "string"
+            ? new Ed25519PublicKey(this._account.publicKey)
```
GhostWalker562, line 656: Is it possible to adjust this a bit so that we can account for SingleKey's as well?

If an outdated wallet shares any other public key that is not Ed25519, this will probably break

packages/wallet-adapter-core/src/LegacyWalletPlugins/types.ts
```diff
@@ -74,7 +79,7 @@ export declare interface WalletCoreEvents {
   readyStateChange(wallet: Wallet): void;
   standardWalletsAdded(wallets: Wallet | AptosStandardSupportedWallet): void;
   networkChange(network: NetworkInfo | null): void;
-  accountChange(account: AccountInfo | null): void;
+  accountChange(account: AccountInfo | StandardAccountInfoInput | null): void;
```
GhostWalker562, line 82: Do you think it might be better to just share the `AccountInfoOutput` instead?

0xmaayan: This is tricky as the adapter still supports legacy and new standard types, so the account the adapter gets back from the wallet can be of both types. Changing it gets us into a type-rabbit-hole, so we simply change the actual input type the adapter returns.

In the near future, we should probably support _only_ AIP-62 wallets

packages/wallet-adapter-core/src/WalletCore.ts
0xmaayan: I'd argue, if you are an outdated wallet - you should update.

Most of the used / updated wallets are already on AIP-62 so they should never really get here, we keep it for the small amount of wallets the adapter still needs to support - which we should probably deprecate in the near future.

In the adapter context, if the `public key` is of type string, we assume it is an old standard which most chances is just Ed255519.

GhostWalker562 (approved):

packages/wallet-adapter-core/src/WalletCore.ts
hardsetting: Agreed, the old interface does not support non-ed25519 public keys, so if they were serializing other types of keys they were probably doing something wrong

packages/wallet-adapter-core/src/LegacyWalletPlugins/types.ts
hardsetting: I think we should use `AccountInfo` from `@aptos-labs/wallet-standard` here, that uses the `PublicKey` type.

If I understand correctly, this is an event that can be listened to externally by the dapp.
Since we're bumping the major, we might as well clean up the interface and have the "right" type here.

In case of legacy plugins, we should be able to adapt the return types to `AccountInfo` so that the dapp sees one consistent return type.

EDIT: [here's](https://github.com/aptos-labs/aptos-wallet-adapter/blob/bde811255e523d050844885312cc2f2e0d73c91d/packages/wallet-adapter-core/src/WalletCore.ts#L1139) a good spot where we could normalize

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -131,7 +131,7 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
   private _wallet: Wallet | null = null;
 
   // Current connected account
-  private _account: AccountInfo | null = null;
+  private _account: AccountInfo | StandardAccountInfoInput | null = null;
```
hardsetting, line 134: Same here, let's just use `AccountInfo` from the standard. We can convert it to the right type before setting its value

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -465,7 +465,7 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
    * @param account An account
    */
   private ensureAccountExists(
-    account: AccountInfo | null
+    account: AccountInfo | StandardAccountInfoInput | null
```
hardsetting, line 468: same here

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -541,57 +541,12 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
-        return;
-      }
-    }
-
-    // account is of type `AccountInfo`
-    this._account = { ...(account as AccountInfo) };
-    return;
+  setAccount(account: AccountInfo | StandardAccountInfo | null): void {
+    this._account = account;
   }
```
hardsetting, line 550: This would be a good spot where we support both `AccountInfo`s (the old internal type, and the standard type) as inputs, and normalize it to the standard one.

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -687,9 +642,30 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
    * @return account info
    * @throws WalletAccountError
    */
-  get account(): AccountInfo | null {
+  get account(): AccountInfoOutput | null {
```
hardsetting, line 645: Let's just return `AccountInfo` from the wallet standard package

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -687,9 +642,30 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
+        publicKey:
+          // for backward compatibility, if the account public key is of type `string` convert it to Ed25519PublicKey
+          typeof this._account.publicKey === "string"
+            ? new Ed25519PublicKey(this._account.publicKey)
+            : // for backward compatibility, if the account public key is of type `string[]` convert each string to Ed25519PublicKey
+              Array.isArray(this._account.publicKey) &&
+                this._account.publicKey.every(
+                  (item) => typeof item === "string"
+                )
+              ? this._account.publicKey.map((key) => new Ed25519PublicKey(key))
```
hardsetting, line 662: Ideally we should convert `string[]` and `minKeysRequired` to `MultiEd25519PublicKey` that extends `PublicKey`

apps/nextjs-example/src/app/page.tsx
```diff
@@ -147,7 +150,7 @@ function WalletSelection() {
 }
 
 interface WalletConnectionProps {
-  account: AccountInfo | null;
+  account: AccountInfoOutput | null;
```
hardsetting, line 153: let's use `AccountInfo` from the standard

packages/wallet-adapter-core/src/utils/walletSelector.ts
```diff
@@ -44,9 +45,9 @@ export function isInstallRequired(wallet: AnyAptosWallet) {
 }
 
 /** Truncates the provided wallet address at the middle with an ellipsis. */
-export function truncateAddress(address: string | undefined) {
+export function truncateAddress(address: AccountAddress | undefined) {
```
hardsetting, line 48: small nit: we could use `AccountAddressInput`, but nbd

hardsetting (changes requested):
Thanks for making this change!
As per my comments, I think we should try using `AccountInfo` from the standard :)

hardsetting:
By quickly looking at the code, I think we could do this:
- define a `normalizeAccountInfo` that normalizes any `AccountInfo` to the standard's one. Here you would put the code that converts `string` to `Ed25519PublicKey` and `(string[], minKeysRequired)` to `MultiEd25519PublicKey`
- when calling `connectWallet` [here](https://github.com/aptos-labs/aptos-wallet-adapter/blob/bde811255e523d050844885312cc2f2e0d73c91d/packages/wallet-adapter-core/src/WalletCore.ts#L781-L786), we normalize the account info before calling `setAccount`
- when we receive an "accountChange" event [here](https://github.com/aptos-labs/aptos-wallet-adapter/blob/bde811255e523d050844885312cc2f2e0d73c91d/packages/wallet-adapter-core/src/WalletCore.ts#L1139), we normalize the account info before `setAccount` and `emit`
- `setAccount` should only take the standard's `AccountInfo` as input

basically.. let's normalize as soon as we have a chance to make our life easier :)

0xmaayan:
closed in favor of https://github.com/aptos-labs/aptos-wallet-adapter/pull/464

PR #456, opened 2024-12-02
Support network based api keys

0xmaayan:
We recently added API Key support to the adapter, but an API key is a network based, so supporting only 1 api key doesnt make sense when the user connects their wallet to different networks.

This PR adds support for a network based API Key so the adapter uses the correct API Key based on the network the wallet is connected to

### DIFF
--- a/packages/wallet-adapter-core/src/utils/helpers.ts
+++ b/packages/wallet-adapter-core/src/utils/helpers.ts
@@ -80,16 +80,24 @@ export const getAptosConfig = (
   if (!networkInfo) {
     throw new Error("Undefined network");
   }
+  const currentNetwork = convertNetwork(networkInfo);
+
+  if (isAptosLiveNetwork(currentNetwork)) {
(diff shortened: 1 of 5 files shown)


kent-white (approved):

packages/wallet-adapter-core/src/utils/helpers.ts
```diff
@@ -108,6 +116,17 @@ export const isAptosNetwork = (
   return NetworkToNodeAPI[networkInfo.name] !== undefined;
 };
 
+export const isAptosLiveNetwork = (networkInfo: Network): boolean => {
+  if (!networkInfo) {
```
hardsetting: did you encounter this case?

hardsetting (approved):

packages/wallet-adapter-core/src/utils/helpers.ts
0xmaayan: oops, copy-past, removed

PR #467, opened 2025-02-01
Fix unexpected exceptions when submitting txn with a CUSTOM network

SamuelQZQ:
The `convertNetwork` cannot be called with CUSTOM network. So we change `getAptosConfig` to only called it when it's aptos netowrk.

### DIFF
--- a/packages/wallet-adapter-core/src/utils/helpers.ts
+++ b/packages/wallet-adapter-core/src/utils/helpers.ts
@@ -80,25 +80,27 @@ export const getAptosConfig = (
   if (!networkInfo) {
     throw new Error("Undefined network");
   }
-  const currentNetwork = convertNetwork(networkInfo);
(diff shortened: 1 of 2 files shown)


packages/wallet-adapter-core/src/utils/helpers.ts
```diff
@@ -80,25 +80,27 @@ export const getAptosConfig = (
...
+    }
+
     return new AptosConfig({
       network: currentNetwork,
-      clientConfig: { API_KEY: apiKey ? apiKey[currentNetwork] : undefined },
     });
-  }
-
-  if (isAptosNetwork(networkInfo)) {
+  } else {
```
0xmaayan: nit: unnecessary else

0xmaayan (approved):

PR #464, opened 2025-01-15
Upgrade to AIP-62 standard

0xmaayan:
This PR migrates the wallet adapter to support only the AIP-62 format and compatible wallets.

The AIP-62 standard is a chain-agnostic set of interfaces and conventions designed to enhance how applications interact with injected wallets. See more https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-62.md

By migrating to AIP-62, we can deprecate unmaintained or unused Aptos wallets, ensuring support is provided only for actively developed and functional wallets.

Additionally, dApps will be able to fetch and utilize wallet type information—a feature that has been highly requested by both the community and internal developers, see https://aptos-org.slack.com/archives/C06TWPH09Q9/p1721771598556349)

Live demo https://aptos-labs.github.io/aptos-wallet-adapter/nextjs-example-testing/

PR #473, opened 2025-02-05
New wallet adapter interface

hardsetting:
My proposal for a new interface of the Aptos Wallet Adapter.

Key objectives:
- prioritize newer features (keep interface as close as possible to wallet standard)
- stay backward compatible in a way that doesn't bloat the "happy path"
- reduce the amount of wrapper code
- ensure the state of the adapter stays consistent
- simplify the developer's life

Changes in `core` package:
- Defined `AdaptedWallet` class, that will take as an input a wallet interface with minimum required features and expose a unified and clean interface for accessing it. Internally, this class performs feature version negotiation, ensuring we're using the most up-to-date version or fallback to older but still compatible versions.
- Defined a structured way to access registered wallets (using `WalletAdapter`).
- This package will not deal with persisted state e.g. window.localStorage. This will be the responsibility of more specialized packages
- Create utility conversion function `walletFromLegacyPlugin` so that dapps that rely on legacy plugins can still use the new interface (all required features need to be supported)

Changes in `react` package:
- Provide "reactive" interface to aptos wallets.
- Define hooks that listen to events and keep values up-to-date at all times (ensures state consistency)
- Intended for web-based dapps (might extract web logic into separate package if there's value)
- Can be used to build custom wallet selectors
- Defines the concept of "active wallet" and the logic to persist its state
- The value returned by the hook `useActiveWallet` has different states. Union discrimination enables intuitive and expressive access to features based on the current state.

example:
```tsx
function Navbar() {
  const wallet = useActiveWallet();
  return (
    <nav>
      <div>Logo</div>
      {wallet.isConnected ? <ActiveAccountNavItem wallet={wallet} /> : <WalletSelector />}
    </nav>
  );
}

// Requires connected wallet as input
export function ActiveAccountNavItem({ wallet }: { wallet: ConnectedWallet }) {
  const [isDropdownOpen, setIsDropdownOpen] = useState(false);

  const onToggle = () => {
    setIsDropdownOpen((prev) => !prev);
  };

  const onDisconnect = async () => {
    // only available if the wallet is connected
    await wallet.disconnect();
  };

  return (
    <button onClick={onToggle}>
      <div>{wallet.activeAccount.address.toString()}</div>
      {isDropdownOpen ? (
        <ul>
          <li>Copy address</li>
          <li onClick={onDisconnect}>Disconnect</li>
        </ul>
      ) : null}
    </button>
  )
}
```

TODO:
- [x] [ui] build on top of the `react` package and showcase it's expressivenesss
- [x] [core] bring back analytics and formalize events
- [x] [core] bring back opting out of certain wallets
- [ ] [react/ui] figure out how to approach "sdk" wallets. should we pass them during context providing? should we just register them at startup like extension wallets?
- [ ] [ui] bring back "opt-in" wallets and explore universal links


packages/wallet-adapter-react/src/new/useActiveWallet.ts
```diff
@@ -0,0 +1,115 @@
...
+  account?: undefined;
+  network?: undefined;
+  wallet?: undefined;
+}
+
+export interface UseActiveWalletConnectedResult {
+  isInitialized: true;
+  isConnected: true;
+  account: AccountInfo;
+  network: Network;
```
hardsetting: maybe rename "activeAccount and "activeNetwork"?

packages/wallet-adapter-core/src/new/AdaptedWallet.ts
```diff
@@ -0,0 +1,254 @@
...
+
+export interface AdaptedWalletConfig {
+  disableTelemetry?: boolean;
+}
+
+/**
+ * A wallet instance adapted from an Aptos standard wallet that supports
+ * all required features with minimum version.
+ */
+export class AdaptedWallet {
```
0xmaayan: This can probably live in the wallet-standard repo, I actually have a long standing open PR for that https://github.com/aptos-labs/wallet-standard/pull/7

packages/wallet-adapter-core/src/new/AdaptedWallet.ts
```diff
@@ -0,0 +1,254 @@
...
+  readonly name: string;
+  readonly url: string;
+  readonly icon: WalletIcon;
+  readonly features: AptosFeatures;
+
+  readonly availableNetworks: Network[];
+
+  // Google Analytics 4 module
+  private readonly ga4?: GA4;
```
0xmaayan: why this needs to be part of the AdapterWallet class? it is used for the wallet-core product tracking

packages/wallet-adapter-core/src/new/AdaptedWallet.ts
```diff
@@ -0,0 +1,254 @@
...
+        this.availableNetworks.push(network);
+      }
+    }
+  }
+
+  // TODO: revise event formats and names
+  private recordEvent(eventName: string, additionalInfo?: object) {
+    this.ga4?.gtag("event", `wallet_adapter_${eventName}`, {
+      wallet: this.name,
+      // network: this._network?.name,
```
0xmaayan: why comment? it is actually very helpful for our metrics

packages/wallet-adapter-core/src/new/AdaptedWallet.ts
```diff
@@ -0,0 +1,254 @@
...
+  }
+
+  // endregion
+
+  // region Accounts
+
+  readonly onAccountConnectedListeners = new Set<(account: AccountInfo) => void>();
+  readonly onAccountDisconnectedListeners = new Set<() => void>();
+
+  async getConnectedAccounts(): Promise<AccountInfo[]> {
```
0xmaayan: why do we want to represent it as an array? I believe aptos connect is the only wallet that supports multiple accounts?

packages/wallet-adapter-core/src/new/AptosStandardWallet.ts
```diff
@@ -0,0 +1,52 @@
...
+} from '@aptos-labs/wallet-standard';
+
+type FeatureVersion = `${number}.${number}` | `${number}.${number}.${number}`;
+type TargetVersion = `${number}.${number}`;
+
+/**
+ * Required features with minimum versions.
+ * In the future, we might choose to slowly deprecate older versions to simplify the adapter's code.
+ */
+const requiredFeatures: [name: keyof MinimallyRequiredFeatures, version: TargetVersion][] = [
```
0xmaayan, line 15: maybe put in the wallet-standard repo, to ease on maintenance

packages/wallet-adapter-core/src/new/AptosStandardWallet.ts
```diff
@@ -0,0 +1,52 @@
...
+  return major === tgtMajor && minor >= tgtMinor;
+}
+
+/**
+ * Check whether a generic wallet is an Aptos standard wallet.
+ *
+ * The wallet needs to implement all the required features with minimum version.
+ * @param wallet generic wallet to be considered compatible.
+ */
+export function isAptosStandardWallet(wallet: Wallet): wallet is AptosStandardWallet {
```
0xmaayan, line 41: we kind of have it already in the standard repo https://github.com/aptos-labs/wallet-standard/blob/main/src/detect.ts#L27

packages/wallet-adapter-core/src/new/WalletAdapter.ts
```diff
@@ -0,0 +1,117 @@
...
+const finalizationRegistry = supportsFinalizationRegistry
+  ? new FinalizationRegistry<CleanupCallback>((cleanup) => cleanup())
+  : undefined;
+
+export type AdaptedWalletEventCallback = (wallet: AdaptedWallet) => unknown;
+export type UnsubscribeCallback = () => void;
+
+export interface WalletAdapterConfig {
+  disableTelemetry?: boolean;
+  whitelist?: string[];
```
0xmaayan: what are these lists for?

packages/wallet-adapter-core/src/new/utils/walletFromLegacyPlugin.ts
```diff
@@ -0,0 +1,264 @@
+import {
```
0xmaayan: Part of me really wants to stop legacy plugin supports, at this point we have been given enough time for wallets to migrate and all major wallets are now on AIP-62. Our metrics also show that.

Deprecate legacy wallets will also make sure no new wallets coming to the ecosystem implemented with legacy flow in mind.

See more context https://aptos-org.slack.com/archives/C06TWPH09Q9/p1738169671855909?thread_ts=1721771598.556349&cid=C06TWPH09Q9

packages/wallet-adapter-react/src/new/useActiveWallet.ts
```diff
@@ -0,0 +1,122 @@
...
+  AptosSignMessageInput,
+  AptosSignTransactionInputV1_1,
+  WalletIcon,
+} from '@aptos-labs/wallet-standard';
+import { useEffect, useMemo, useState } from 'react';
+import { useActiveWalletId } from './useActiveWalletId';
+import { useAvailableWallets } from './useAvailableWallets';
+
+export interface UninitializedWallet {
+  isInitialized: false;
```
0xmaayan, line 15: what does initialize/uninitialize mean?

packages/wallet-adapter-react/src/new/useActiveWallet.ts
```diff
@@ -0,0 +1,122 @@
...
+  isInitialized: true;
+  isConnected: true;
+  activeAccount: AccountInfo;
+  activeNetwork: Network;
+  name: string;
+  icon: WalletIcon;
+  features: AptosFeatures;
+  disconnect: () => Promise<void>;
+  signMessage: (input: AptosSignMessageInput)
+    => ReturnType<AdaptedWallet['signMessage']>;
```
0xmaayan: TIL

packages/wallet-adapter-react/src/new/useActiveWalletId.ts
```diff
@@ -0,0 +1,30 @@
+import { useEffect, useState } from 'react';
+
+const activeWalletIdStorageKey = '@aptos-labs/wallet-adapter/active-wallet';
+
+export function getActiveWalletId() {
+  return window.localStorage.getItem(activeWalletIdStorageKey) ?? undefined;
+}
+
+export function setActiveWalletId(walletId: string | undefined) {
```
0xmaayan, line 13: `walletId` can probably just be the wallet name

0xmaayan (commented):
Nice refactoring and thinking. How hard would it be to rebase it with https://github.com/aptos-labs/aptos-wallet-adapter/pull/464 ?

Some Qs
- I dont see any mobile specific handling
- I feel some implementations can easily live in the standard repo, WalletAdapter interface for example
- We need to think how these changes go with other workstream we have for the cross chain adapter

hardsetting:
> - I dont see any mobile specific handling

true, mobile specific handling and registry wallets have not been implemented yet

> - I feel some implementations can easily live in the standard repo, WalletAdapter interface for example

Yes that's true. I added the code here to keep the changes in one spot, but it can be under the standard's package.
On this topic, I was actually wondering why don't we move the standard's package in this monorepo?
This way the wallet adapter can be more closely tied to the standard.
What do you think?

> - We need to think how these changes go with other workstream we have for the cross chain adapter

Yes agreed, if you have any doc or ongoing conversations I'd be happy to take a look to make sure these changes are compatible and take into account the cross-chain adapter

hardsetting:
> How hard would it be to rebase it with https://github.com/aptos-labs/aptos-wallet-adapter/pull/464 ?

Not hard at all!
I kept all changes under the "new" subpackage for now so there's basically no overlap

0xmaayan:
Writing here my thoughts and post-discussion agreements
- Wallet Core should only be responsible on 
  - defining interfaces and types 
  - logic to detect extension wallets (aip-62 compatible)
  - manage and append SDK wallets
  - We should make sure we dont break external packages using the Wallet Core directly, i.e [wallet aggregator packages](https://github.com/XLabs/wallet-aggregator-sdk/tree/master/packages/wallets/aptos ) and projects that dont use the React provider. That means detect wallets and wallet functions like connect, disconnect, signAndSubmitTransaction, etc
  - Should support configurations like optin/optout wallets, network to support (for SDK wallets like AC) and other SDK specific wallets config, as well as disable telementry.
- Wallet (React) Provider should be responsible on the dapp state (connected wallet, current account, etc) and run wallet actions like connect, disconnect, sign transaction, sign and submit transaction, etc). It should also be configurable with the wallet core configs as dapps probably mainly use this package.
- Wallet Selector UI, we should deprecate the Antd package, and mainly support the MUI package (explorer and ANS use it). As we introduce the shadcn/ui selector, dapps can take full control on the provided wallet selector although most of the dapps use their own custom selectors.

This changes can be then combined with other workstream we have for the wallet adapter and be released with a major upgrade

0xmaayan:
> Yes that's true. I added the code here to keep the changes in one spot, but it can be under the standard's package.
> On this topic, I was actually wondering why don't we move the standard's package in this monorepo?
> This way the wallet adapter can be more closely tied to the standard.
> What do you think?

Agreed!

packages/wallet-adapter-core/src/new/AdaptedWallet.ts
hardsetting: Happy to move it elsewhere! Where were you thinking of putting this?

packages/wallet-adapter-core/src/new/AdaptedWallet.ts
hardsetting: We don't have direct access to the active network here at the moment.
I can add it, it just adds a bit of complexity.

packages/wallet-adapter-core/src/new/AdaptedWallet.ts
hardsetting: I think that more generally, we should distinguish between the concept of "connected account" and "active account".

At the moment, Petra only supports requesting signatures for the active account, but technically nothing prevents the dapp from requesting a signature by a different connected account (or even network).
We just decided that it's "safer" this way.

As you mentioned, Aptos Connect allows specifying the signer address and network, and I want to be able to allow wallets to do that.

packages/wallet-adapter-core/src/new/AptosStandardWallet.ts
hardsetting: True, I just wanted to improve the logic without having to make a new release in the wallet-standard repo.
We should eventually move it over.

packages/wallet-adapter-core/src/new/WalletAdapter.ts
hardsetting: By default, all wallets implementing the standard will be considered.
These list allow the dapp developer to selectively ignore unwanted wallets, or explicitly defining which wallets are supported.
The check is done by name

packages/wallet-adapter-core/src/new/utils/walletFromLegacyPlugin.ts
hardsetting: Yeah I support this. I wouldn't mind removing this file altogether.
It's a nice example on how to adapt other objects into standard wallets. (e.g. cross-chain wallets)

packages/wallet-adapter-react/src/new/useActiveWallet.ts
hardsetting: In order for the wallet to be "usable" aka "initialized", we have to fetch some information e.g. `activeAccount` and `activeNetwork`. This step is async so the wallet is going to be in a "loading" state for a couple frames.

The possible states are:
- uninitialized (we can potentially show a spinner in the UI)
- intialized, not connected
- initialized, connected

We can add a caching logic later, but figured we should keep it simple for now :)

packages/wallet-adapter-react/src/new/useActiveWalletId.ts
hardsetting: Yes, at the moment the name is effectively the identifier.
I was hoping to differentiate between "id" and "display name", so we can have:

```typescript
{ id: 'petra', name: "Petra", ... }
{ id: 'aptos-connect-google', name: "Continue with Google", ... }
{ id: 'aptos-connect-apple', name: "Continue with Apple", ... }
```

but I also recognize there isn't a functional reason to differentiate at the moment.
It just "feels" right haha

packages/wallet-adapter-core/src/new/AdaptedWallet.ts
0xmaayan: hmm this is tricky, the purpose of it is to record wallet events like connect, disconnect, signAndSubmitTransaction, etc.

So it makes sense to keep it here and record each `AdaptedWallet` method. 
But it also means we create a `ga4` instance for each `AdaptedWallet` https://github.com/aptos-labs/aptos-wallet-adapter/pull/473/files#diff-f9f367b57dfbc51263db65e5ad66417c80df2a0a83a21200c5a1eeb48988d3ceR39 and that feels unnecessary.

Maybe we can initiate it in the `WalletAdapter` class, and pass it to the `AdaptedWallet` instance to be used in each wallet function

packages/wallet-adapter-core/src/new/AdaptedWallet.ts
0xmaayan: hmm we have it in the provider `useActiveWallet`, maybe we can get it from there?

packages/wallet-adapter-core/src/new/AptosStandardWallet.ts
0xmaayan: gotcha, lets leave a comment then, and move it there before releasing it

packages/wallet-adapter-core/src/new/WalletAdapter.ts
0xmaayan: Maybe we can use a stronger type? something like what we have now
```
type AvailableWallets = "Petra" | "Pontem" | "Nightly" | ....

whitelist?: AvailableWallets[]
```

packages/wallet-adapter-core/src/new/network.ts
```diff
@@ -0,0 +1,28 @@
...
+  MAINNET: AptosNetwork.MAINNET,
+  TESTNET: AptosNetwork.TESTNET,
+  DEVNET: AptosNetwork.DEVNET,
+  LOCAL: AptosNetwork.LOCAL,
+} as const;
+
+/**
+ * Custom network configuration. Requires at least the full node URL to work properly.
+ */
+export interface CustomNetwork {
```
0xmaayan, line 19: this should be aligned with the `AptosConfig` instance in the sdk, right?
https://github.com/aptos-labs/aptos-ts-sdk/blob/main/tests/unit/aptosConfig.test.ts#L71

packages/wallet-adapter-core/src/new/AdaptedWallet.ts
```diff
@@ -0,0 +1,254 @@
...
+    return feature.signMessage(input);
+  }
+
+  async signTransaction(input: AptosSignTransactionInputV1_1) {
+    const feature = this.features['aptos:signTransaction']
+
+    if (isFeatureMinorVersion(feature, "1.0")) {
+      const { signerAddress, feePayer } = input;
+      // This will throw an error if it requires an async call
+      const transaction = buildTransaction(input);
```
0xmaayan: hmm looks like we are now requiring the dapp to provide the account sequence number?

packages/wallet-adapter-core/src/new/utils/version.ts
```diff
@@ -0,0 +1,10 @@
+export function isFeatureMinorVersion<
```
0xmaayan, line 1: i'd move it to the standard repo

packages/wallet-adapter-core/src/new/utils/walletFromLegacyPlugin.ts
0xmaayan: interesting.... then maybe we can put it under an `example` folder? I think for cross-chain wallets we will probably manage it in the cross-chain package

packages/wallet-adapter-core/src/new/network.ts
hardsetting: Yes, ideally

I wanted to make the network definition a bit more expressive i.e.
- when using standard networks you just pass the network name as shorthand, everything else will be inferred for you
- when using a custom network, you have to pass it in object form, and some fields are required

at the moment, in the SDK everything other than `network` is optional so you can pass a standard network with overrides, or a custom network with missing fields

if someone is using `mainnet`, `testnet` ... I don't think there's a big use-case for overriding values.
That's what a "custom" network would be. But we can still allow that if needed

packages/wallet-adapter-core/src/new/AdaptedWallet.ts
hardsetting: `signAndSubmitTransaction@1.0` doesn't support payload inputs as arguments.
If the wallet doesn't implement `signAndSubmitTransaction@1.1` and the dapp is passing a payload input, we try our best to make it work.

Better to have the wallet adapter give you an expressive and actionable error, than having the popup not opening on safari :D

I also thought of allowing wallets to specify whether they're ok with async calls through a flag.
For example, chrome extensions don't have that limitations. Aptos Connect does.

0xmaayan: > I also thought of allowing wallets to specify whether they're ok with async calls through a flag.
> For example, chrome extensions don't have that limitations. Aptos Connect does.

That could be interesting - I assume since AC has this limitation, other SDK wallets also have it

packages/wallet-adapter-core/src/new/network.ts
0xmaayan: > when using a custom network, you have to pass it in object form, and some fields are required

Yep, but dont we want to have then pass in an object similar to the one that will eventually be used with the SDK to initiate an Aptos connection?

0xmaayan (commented):
Does this change handle the different SDK wallets we currently support (mizu, t wallet)? I can't find it. 
https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-core/src/sdkWallets.ts

packages/wallet-adapter-core/src/new/registry.ts
```diff
@@ -0,0 +1,62 @@
...
+ */
+export function makeDefaultRegistry({ aptosConnect = {} }: DefaultRegistryConfig = {}): RegistryWallet[] {
+  const isSSR = typeof window === "undefined";
+  const aptosConnectWallets = !isSSR && aptosConnect ? [
+    new AptosConnectGoogleWallet(aptosConnect),
+    new AptosConnectAppleWallet(aptosConnect),
+  ] : [];
+  return [
+    ...aptosConnectWallets,
+    petra,
```
0xmaayan, line 58: we are missing rimosafe and okx
https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-core/src/registry.ts#L44-L55

packages/wallet-adapter-core/src/new/registry.ts
```diff
@@ -0,0 +1,62 @@
...
+export type RegistryWallet = DiscoverableWallet | AptosStandardWallet;
+
+/**
+ * Default registry of AIP-62 wallet standard wallets.
+ * This list is used to show supported wallets even if they are not installed on the user machine.
+ */
+export function makeDefaultRegistry({ aptosConnect = {} }: DefaultRegistryConfig = {}): RegistryWallet[] {
+  const isSSR = typeof window === "undefined";
+  const aptosConnectWallets = !isSSR && aptosConnect ? [
+    new AptosConnectGoogleWallet(aptosConnect),
```
0xmaayan, line 53: on the UI, the wallet selector differentiate between aptosConnectWallets and installedWallets and notDetectedWallets,
This is to then show those 3 different sections with a different UI 
https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-core/src/utils/walletSelector.ts#L119

Might worth to have 3 different functions for each wallet type so the UI can easily design it however it wants

0xmaayan (commented):
I think it looks good, but need to think on the migration plan so we wont break users that use this class directly or their own provider.

PR #475, opened 2025-02-07
Custom network fix

gregnazario:


hardsetting:
Would it be complex to having the formatting committ in a separate PR?

packages/wallet-adapter-core/src/utils/helpers.ts
```diff
@@ -97,10 +97,9 @@ export const getAptosConfig = (
     });
   }
 
-  return new AptosConfig({
-    network: Network.CUSTOM,
-    fullnode: networkInfo.url,
-  });
+  throw new Error(
+    "Invalid network, custom network not supported with Aptos wallet adapter to prevent user from using an unexpected network.",
```
banool, line 102: Maybe leave a comment telling the dev what they should do configuration-wise to avoid this issue.

banool (approved):
#movefast 

It'd be good to have CI that makes sure it stays formatted as intended (we can add that in another PR)

PR #482, opened 2025-02-19
[core][react] Add aptos:signIn feature

GhostWalker562:
### Description

Added support for the new `aptos:signIn` feature in the updated wallet standard.

### Tests

https://github.com/user-attachments/assets/40b3bbc6-4ea4-450a-8575-5eb669c7e91e

### DIFF
--- a/.changeset/clean-kings-battle.md
+++ b/.changeset/clean-kings-battle.md
@@ -1,0 +1,6 @@
+---
+"@aptos-labs/wallet-adapter-react": minor
+"@aptos-labs/wallet-adapter-core": minor
+---
+
+Add support for aptos:signIn feature
--- a/packages/wallet-adapter-core/src/WalletCore.ts
+++ b/packages/wallet-adapter-core/src/WalletCore.ts
@@ -528,8 +528,66 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
       }
     }
 
-    // Now we can connect to the wallet
-    await this.connectWallet(selectedWallet);
+    // If true, the wallet was redirected to the mobile app's browser.
+    if (this.redirectIfRedirectable(selectedWallet)) return;
+
+    await this.connectWallet(selectedWallet, async () => {
+      const response = await selectedWallet.features["aptos:connect"].connect();
+      if (response.status === UserResponseStatus.REJECTED) {
+        throw new WalletConnectionError("User has rejected the request")
+          .message;
+      }
+
+      return { account: response.args, output: undefined };
+    });
+  }
+
+  /**
+   * Signs into the wallet by connecting and signing an authentication messages.
+   *
+   * For more information, visit: https://siwa.aptos.dev
+   *
+   * @param args
+   * @param args.input The AptosSignInInput which defines how the SIWA Message should be constructed
+   * @param args.walletName The name of the wallet to sign into
+   * @returns The AptosSignInOutput which contains the account and signature information
+   */
+  async signIn(args: {
+    input: AptosSignInInput;
+    walletName: string;
+  }): Promise<AptosSignInOutput> {
+    const { input, walletName } = args;
+
+    const allDetectedWallets = this._standard_wallets;
+    const selectedWallet = allDetectedWallets.find(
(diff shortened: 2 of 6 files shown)


.changeset/clean-kings-battle.md
```diff
@@ -0,0 +1,6 @@
+---
+"@aptos-labs/wallet-adapter-react": patch
+"@aptos-labs/wallet-adapter-core": patch
```
hardsetting: I think this should be a minor change?

hardsetting (approved):
Looks great! Just a comment about version bump

.changeset/clean-kings-battle.md
GhostWalker562: I'll change it 👍

kent-white (approved):

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -564,6 +568,91 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
     }
   }
 
+  async signIn(
```
0xmaayan: If it makes sense, please use JSON input here, I moved other functions to use it where it makes sense

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -564,6 +568,91 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
+      throw new WalletNotSupportedMethod(
+        `aptos:signIn is not supported by ${walletName}`
+      ).message;
+    }
+
+    // Check if we are in a redirectable view (i.e on mobile AND not in an in-app browser)
+    // If we are, navigate to the mobile app's browser instead.
+    //
+    // Ignore if wallet is installed (iOS extension)
+    if (isRedirectable()) {
```
0xmaayan, line 636: might just move it to its own function instead of duplicate code

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -564,6 +568,91 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
+          window.location.href = location;
+          return;
+        }
+      }
+    }
+
+    return this.signInWallet(input, selectedWallet);
+  }
+
+  async signInWallet(
```
0xmaayan: ok we should definitely extract the duplicate logic here and in the `connect` method

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -564,6 +568,91 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
+    input: AptosSignInInput,
+    selectedWallet: AdapterWallet
+  ): Promise<void | AptosSignInOutput> {
```
0xmaayan: dont think we need `void`
```suggestion
  ): Promise<AptosSignInOutput> {
```

GhostWalker562: The issue is that when we redirect to the mobile app, we can't really return anything since there's no output. I could possibly remove the redirect behavior since the flow should be...

1. Attempt to `signIn`
2. If not supported, attempt to regular `connect`
3. Regular `connect` will redirect to mobile app

packages/wallet-adapter-core/src/WalletCore.ts
GhostWalker562: Refactored `isRedirectable` and `signInWallet`, please take a look :)

packages/wallet-adapter-core/src/WalletCore.ts
GhostWalker562: Refactored isRedirectable and signInWallet, please take a look :)

0xmaayan (approved):

packages/wallet-adapter-core/src/WalletCore.ts
GhostWalker562: Removed from the flow

PR #483, opened 2025-02-19
[react] Add support for async functions in autoConnect

GhostWalker562:
### Description

Allow asynchronous functions to determine whether a `WalletProvider` should attempt to `autoConnect` this enable uses cases such as "Sign in with Aptos" which may not want to traditionally connect using the `connect` method.

### Tests

https://github.com/user-attachments/assets/9104086f-33ce-430e-b01b-307c5ca86a4d

### DIFF
--- a/packages/wallet-adapter-react/src/WalletProvider.tsx
+++ b/packages/wallet-adapter-react/src/WalletProvider.tsx
@@ -16,16 +16,19 @@ import {
   Network,
   InputSubmitTransactionData,
   PendingTransactionResponse,
+  WalletReadyState,
   AptosSignInInput,
   AptosSignInOutput,
 } from "@aptos-labs/wallet-adapter-core";
-import { ReactNode, FC, useState, useEffect, useCallback } from "react";
+import { ReactNode, FC, useState, useEffect, useCallback, useRef } from "react";
 import { WalletContext } from "./useWallet";
 
 export interface AptosWalletProviderProps {
   children: ReactNode;
   optInWallets?: ReadonlyArray<AvailableWallets>;
-  autoConnect?: boolean;
+  autoConnect?:
+    | boolean
+    | ((core: WalletCore, adapter: AdapterWallet) => Promise<boolean>);
   dappConfig?: DappConfig;
   disableTelemetry?: boolean;
   onError?: (error: any) => void;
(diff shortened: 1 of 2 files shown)


packages/wallet-adapter-react/src/WalletProvider.tsx
```diff
@@ -76,13 +81,62 @@ export const AptosWalletAdapterProvider: FC<AptosWalletProviderProps> = ({
...
-        connect(localStorage.getItem("AptosWalletName") as string);
-      } else {
-        // if we dont use autoconnect set the connect is loading to false
-        setIsLoading(false);
-      }
+    // Only attempt to auto connect once per render and only if there are wallets
+    if (didAttemptAutoConnectRef.current || !walletCore?.wallets.length) {
+      return;
+    }
+    didAttemptAutoConnectRef.current = true;
```
hardsetting, line 90: This is never set to false. I'm guessing this is by design but pointing out jic :)

hardsetting (approved):

hardsetting:
should probably base this on #482 ?

packages/wallet-adapter-react/src/WalletProvider.tsx
GhostWalker562: Yep by design, it's initialized as `false` but will never be `false` again. If they want to retry connecting they should use `connect` or `signIn` explicitly instead of needing to connect through `autoConnect` once again.

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -564,6 +568,91 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
     }
   }
 
+  async signIn(
```
0xmaayan: should probably first merge the core changes and then rebase

0xmaayan (commented):
Before releasing, please deploy it to the testing env, with `pnpm run deploy:example-testing` to make sure nothing breaks and everything is working as expected

packages/wallet-adapter-core/src/WalletCore.ts
GhostWalker562: Just changed the base, the diffs in the PR should reflect properly now

PR #498, opened 2025-03-10
Fix mobile redirect logic

0xmaayan:
On mobile, extension wallets are not installed and are in `NotDetected` state. The adapter adds these wallets to the `_standard_not_detected_wallets`, so when users try to connect, we first need to check if we are on mobile view and look for the selected wallet in the `_standard_not_detected_wallets` array and handle the `deeplinkProvider` case. Otherwise, we continue to handle the rest of the wallets that are in "Installed" state, i.e SDK wallets.

### DIFF
--- a/.changeset/two-kiwis-serve.md
+++ b/.changeset/two-kiwis-serve.md
@@ -1,0 +1,6 @@
+---
+"@aptos-labs/wallet-adapter-core": patch
+"@aptos-labs/wallet-adapter-nextjs-example": patch
+---
+
+Fix mobile redirect logic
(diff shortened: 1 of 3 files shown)


gregnazario (approved):
So, this means there will be deeplink buttons for any wallets that support it?

0xmaayan:
> So, this means there will be deeplink buttons for any wallets that support it?

There already is. Any not-detected wallet, that has a deepLink provider, shows up on the wallet selector.
We keep a [registry](https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-core/src/registry.ts) of all not-detected wallets 

We just didnt handle this specific case when introducing the `_standard_not_detected_wallets` https://github.com/aptos-labs/aptos-wallet-adapter/issues/464

PR #500, opened 2025-03-10
Cross chain wallets on Aptos

0xmaayan:
Introducing cross-chain wallets on Aptos functionality, that can be used to build and support a cross-chain dapp on Aptos.

Live example https://aptos-labs.github.io/aptos-wallet-adapter/nextjs-cross-chain-example/

**cross-chain-core**:
A package that holds a cross-chain logic. Currently only provides CCTP transfer logic with Wormhole as the provider. In the Future, we will support more providers (LayerZero, Cellar, etc) and add it to that package.
In addition, this package exposes some related APIs like `getAccountUsdcBalance()`.

**Nextjs cross chain demo dapp**
A demo dapp demonstrates the feasibility of a cross chain wallets functionalities, like CCTP transfer and signing transactions or messages with a wallet.

Future PRs:
- Introduce a cross-chain adapter for easy dapp integration


BriungRi:
You can simplify the nextjs example by having the button for `Sepolia` `Solana` or `Sui` open the connect wallet modal for those wallets

packages/cross-chain-core/src/CrossChainCore.ts
```diff
@@ -0,0 +1,60 @@
...
+} from "./providers/wormhole";
+
+export interface CrossChainDappConfig {
+  network: Network;
+  disableTelemetry?: boolean;
+}
+
+export type AptosAccount = Account;
+
+export type Chain = "Solana" | "Ethereum" | "Sui" | "Aptos";
```
BriungRi: why not enum?

packages/cross-chain-core/src/CrossChainCore.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export type CCTPProviders = "Wormhole";
+
+export type UsdcBalance = {
+  amount: string;
+  decimal: number;
+  display: string;
```
BriungRi: Not sure I would implement `display` in this component. I think better to let the developer choose how they want to display. I think it just gets complicated because

1. Developer might want to only use period to dilineate decimals, while others may want to use comma
2. Developer might want to use some kind of significant figures logic

Just think it's not a good idea to put it in this package because it is hard to undo these decisions later

packages/cross-chain-core/src/CrossChainCore.ts
```diff
@@ -0,0 +1,60 @@
...
+};
+
+export interface CrossChainProvider<
+  TQuoteRequest = any,
```
BriungRi, line 29: is it possible to not use `any`? I'm ok with leaving it as a TODO for this PR

packages/cross-chain-core/src/providers/wormhole/config/mainnet/chains.ts
```diff
@@ -0,0 +1,66 @@
+import { Context } from "../types";
+
+import { ChainsConfig } from "../types";
+
+export const mainnetChains: ChainsConfig = {
```
BriungRi: any reference link to where these values come from? Would be nice to leave as comments

packages/wallet-adapter-aggregator/eip6963/src/index.ts
```diff
@@ -0,0 +1,248 @@
...
+import { convertEip6963WalletToAptosWallet } from "./utils";
+
+export type EIP6963ProviderInfo = {
+  info: {
+    uuid: string;
+    name: string;
+    icon: `data:image/svg+xml;base64,${string}`;
+    rdns: string;
+  };
+  provider: any;
```
hardsetting: uhm?

packages/wallet-adapter-aggregator/core/src/index.ts
```diff
@@ -0,0 +1,73 @@
...
+  ConnectOutput = any,
+  SignMessageInput = any,
+  SignMessageOutput = any,
+  SignTransactionInput = any,
+  SignTransactionOutput = any,
+  OnAccountChangeInput = any,
+  OnNetworkChangeInput = any,
+  SendTransactionInput = any,
+  SendTransactionOutput = any,
+> extends EventEmitter<WalletEvents> {
```
hardsetting: What utility does this wrapper class provide?
Wouldn't it be easier to just map the source chain adapter features into Aptos features?

packages/wallet-adapter-aggregator/eip6963/src/utils.ts
```diff
@@ -0,0 +1,239 @@
...
+        });
+      },
+      version: "1.0.0",
+    },
+  };
+};
+
+const EIP6963_ADDITIONAL_FEATURES = (
+  eip6963Wallet: EIP6963ProviderInfo
+): Eip6963Features => {
```
hardsetting: Why do we need these?
The source wallet already supports `eip6963` features, why are we re-defining them?

packages/wallet-adapter-aggregator/eip6963/src/index.ts
```diff
@@ -0,0 +1,248 @@
...
+  };
+  "eip6963:network": {
+    network: () => Promise<number>;
+    version: string;
+  };
+  // "eip6963:getUsdcBalance": {
+  //   getUsdcBalance: () => Promise<UsdcBalance>;
+  //   version: string;
+  // };
+};
```
hardsetting: I'm a bit confused as to why we need these.
I thought that the goal was to create an abstracted Aptos Account from a source chain.
How are these features needed?

apps/nextjs-example/next.config.mjs
```diff
@@ -2,11 +2,11 @@ const isProd = process.env.NODE_ENV === "production";
 
 /** @type {import('next').NextConfig} */
 const nextConfig = {
-  output: "export",
+  output: "npx serve@latest out",
```
hardsetting: What is this for?

packages/cross-chain-react/src/WalletProvider.tsx
```diff
@@ -0,0 +1,199 @@
...
+    try {
+      if (!provider) {
+        throw new Error("Provider is not set");
+      }
+      const { originChainTxnId, destinationChainTxnId } =
+        await provider.initiateCCTPTransfer({
+          sourceChain,
+          wallet,
+          destinationAddress:
+            "0x38383091fdd9325e0b8ada990c474da8c7f5aa51580b65eb477885b6ce0a36b7",
```
hardsetting: is this intended to be hardcoded?

packages/cross-chain-core/src/providers/wormhole/index.ts
```diff
@@ -0,0 +1,340 @@
...
+      {
+        source: sourceToken,
+        destination: destToken,
+      },
+      sourceContext,
+      destContext
+    );
+
+    const resolver = this._wormholeContext.resolver([
+      routes.CCTPRoute, // manual CCTP
```
hardsetting: wondering what are the other options?
Is there a documentation page for the SDK?

packages/cross-chain-core/src/providers/wormhole/index.ts
```diff
@@ -0,0 +1,340 @@
...
+    ]);
+
+    const route = await resolver.findRoutes(req);
+    const cctpRoute = route[0];
```
hardsetting: assuming first route is always best?

packages/cross-chain-core/src/providers/wormhole/index.ts
hardsetting: can the resolver be cached with the context?
or is it expected to be obtained for each transfer through `context.resolver(...)`?

hardsetting:
Took a good look, great work!
I have some high-level ideas I'd love to chat about when you have time

The flow I was hoping to see was something similar to this (hopefully you have access):
https://excalidraw.com/#json=dtzELJ8ezex0lWN18D-F1,3dKp9B39GvEuf4tifqTOQw

apps/nextjs-example/next.config.mjs
0xmaayan: when starting nextjs build locally `next start`, it does not support `export` anymore

packages/wallet-adapter-aggregator/eip6963/src/index.ts
0xmaayan: Global DAA is still in the exploration path, and it most needed and relevant to a specific dapp we are building which will use a dapp specific DAA implementation.

Still, dapps need a way to communicate with the source wallet default methods, so we need to support it

packages/wallet-adapter-aggregator/eip6963/src/utils.ts
0xmaayan: Standard function types and signature makes it hard to implement source wallet methods as aptos standard methods

packages/wallet-adapter-aggregator/core/src/index.ts
0xmaayan: Standard function types and signature makes it hard to implement source wallet methods as aptos standard methods

packages/wallet-adapter-aggregator/eip6963/src/index.ts
0xmaayan: yea, this should be
```
interface EIP1193Provider {
  on: (event: string, callback: (...args: any[]) => void) => void;
  request: (request: {
    method: string;
    params?: Array<unknown>;
  }) => Promise<any>;
}
```

packages/cross-chain-react/src/WalletProvider.tsx
0xmaayan: just for now, eventually the dapp provides the destination account - either DAA or any other Aptos account

packages/cross-chain-core/src/providers/wormhole/index.ts
0xmaayan: > wondering what are the other options?
For now, Wormhole supports only Manual CCTP for Aptos

0xmaayan: > can the resolver be cached with the context?
can be cached

packages/cross-chain-core/src/providers/wormhole/index.ts
0xmaayan: yes

0xmaayan:
> You can simplify the nextjs example by having the button for `Sepolia` `Solana` or `Sui` open the connect wallet modal for those wallets

That brings up questions around multi chain wallets. For example, Nightly which is a wallet that supports all those chains.
I rather leave it to the dapp to decide how it wants to display the wallets in its own `WalletSelector` implementation

BriungRi (approved):

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -1075,6 +1079,8 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
             }
           }
         } else {
+          console.log("response.args.fullMessage", response.args.fullMessage);
+          console.log("response.args.signature", response.args.signature);
```
hardsetting: nit: remove

pnpm-workspace.yaml
```diff
@@ -1,3 +1,3 @@
 packages:
   - "apps/*"
-  - "packages/*"
+  - "packages/**/*"
```
hardsetting: remove, not needed anymore

turbo.json
```diff
@@ -3,7 +3,9 @@
   "globalEnv": [
     "GAID",
     "NEXT_PUBLIC_APTOS_API_KEY_TESNET",
-    "NEXT_PUBLIC_APTOS_API_KEY_DEVNET"
+    "NEXT_PUBLIC_APTOS_API_KEY_DEVNET",
+    "NEXT_PUBLIC_SWAP_CCTP_MAIN_SIGNER_PRIVATE_KEY",
+    "NEXT_PUBLIC_SWAP_CCTP_SPONSOR_ACCOUNT_PRIVATE_KEY"
```
hardsetting, line 8: are these needed in the global turbo.json? is there any way to move them under the specific app they're used in?

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -354,23 +354,27 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
...
+        const aptos = new Aptos(aptosConfig);
+        const name = await aptos.ans.getPrimaryName({
+          address: this._account.address.toString(),
+        });
 
-      this._account.ansName = name;
+        this._account.ansName = name;
+      } catch (error: any) {
+        console.log(`Error setting ANS name ${error}`);
+      }
```
hardsetting: If i understand correctly, we wrapped with try/catch here?
The error is just console logged, I don't think this adds any value so let's remove it.

Or.. what were you trying to do? Maybe there's some other way

packages/wallet-adapter-core/package.json
```diff
@@ -59,7 +59,7 @@
     "tweetnacl": "^1.0.3"
   },
   "peerDependencies": {
-    "@aptos-labs/ts-sdk": "^1.35.0"
+    "@aptos-labs/ts-sdk": "^1.36.0"
```
hardsetting, line 62: Was there a feature in 1.36.0 that you needed?
Why the upgrade

packages/derived-wallet-ethereum/src/index.ts
```diff
@@ -1,4 +1,5 @@
-export * from './EIP1193DerivedPublicKey';
-export * from './EIP1193DerivedSignature';
-export * from './EIP1193DerivedWallet';
-export * from './setupAutomaticDerivation';
+export * from "./EIP1193DerivedPublicKey";
+export * from "./EIP1193DerivedSignature";
+export * from "./EIP1193DerivedWallet";
+export * from "./setupAutomaticDerivation";
+export * from "./shared";
```
hardsetting, line 5: did you need to use anything from "shared"?

apps/nextjs-example/package.json
```diff
@@ -11,7 +11,7 @@
     "format": "prettier --write './**/*.{ts,tsx,js,jsx,json,md,html,css}'"
   },
   "dependencies": {
-    "@aptos-labs/ts-sdk": "^1.35.0",
+    "@aptos-labs/ts-sdk": "^1.36.0",
```
hardsetting, line 14: was the update needed?

hardsetting (approved):

packages/wallet-adapter-core/src/WalletCore.ts
0xmaayan: It happened when I had a bad connection, and because we dont catch the error it throws and breaks the app. So added a try/catch but I dont want to throw because it is not really an error, I just want the app to not break

packages/wallet-adapter-core/package.json
0xmaayan: yes, `truncateAddress` helper

packages/derived-wallet-ethereum/src/index.ts
0xmaayan: added by mistake, but I think we should still export everything, especially types

apps/nextjs-example/package.json
0xmaayan: see previous comment

turbo.json
0xmaayan: turbo needs to be aware of it when the dapp is deployed to a live server

hardsetting: Yeah I understand, but if you defined them under the example app's `turbo.json` it would also work right?
So it's a little more clearer that they're only related to that specific application.

packages/wallet-adapter-core/src/WalletCore.ts
hardsetting: ok would you mind wrapping only the section around `getPrimaryName` then?

hardsetting:
```typescript
try {
  this._account.ansName = await aptos.ans.getPrimaryName({
    address: this._account.address.toString(),
  });
} catch (err) {
  // either ignore or console log
}
```

packages/derived-wallet-ethereum/src/index.ts
hardsetting: Gotcha!
I had intentionally left it out, it's a utility function for internal use only
but I guess there's no harm done

turbo.json
hardsetting: don't worry about it, we can figure it out later whether that's possible

hardsetting: Just for reference:
https://turbo.build/repo/docs/reference/package-configurations

PR #504, opened 2025-03-17
Derived wallets implementation for Solana and Ethereum

hardsetting:
Implementation of `SolanaDerivedWallet` and `EIP1193DerivedWallet` (Ethereum).

## `derived-wallet-base`
Contains common code used in all derived wallets packages:
- parsing of Aptos signing messages (useful for veryfing messages with abstract public keys)
- format definition and encoding/decoding of StructuredMessage

some of this code could be moved in the wallet-standard repo.

## `derived-wallet-ethereum`
Contains an `EIP1193DerivedWallet` that adapts an `EIP6963ProviderDetail` into a fully functional `AptosWallet`.

See [EIP-1193](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1193.md) (wallet provider) and [EIP-6963](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-6963.md) (wallet discovery) for more information.

It also exports `setupAutomaticEthereumWalletDerivation` that leverages EIP-6963 and the [mipd library](https://github.com/wevm/mipd) for wallet discovery, and registers an `EIP1193DerivedWallet` whenever a new EIP1193-compatible wallet is discovered.

We use [Sign In with Ethereum](https://docs.login.xyz/) (SIWE) for signing messages and transactions, as other message signature methods are not verified by the wallet. We can't show simulations for Aptos transactions on non-Aptos wallets, at the very least we want the dapp's domain and other details to be verified.
A module called `createSiweMessageFromAptos` provides ways to convert a structured message or transaction into a SIWE message that contains human-readable information about the signature input.

## `derived-wallet-solana`
Contains `SolanaDerivedWallet` that adapts a `SolanaWalletAdapter` from `@solana/wallet-standard-wallet-adapter-base` into a fully functional `AptosWallet`.

It also exports `setupAutomaticSolanaWalletDerivation` function that leverages solana's wallet discovery and registers a `SolanaDerivedWallet` whenever a new wallet is detected.

Just like mentioned above for Ethereum, we are using Sign in with Solana (SIWS) for signatures to provide signature input verification, that would otherwise not be possible with the basic `signMessage`. 
A module called `createSiwsMessageFromAptos` provides ways to convert aptos signature inputs to SIWS messages that include human-readable information about the input.

## Some notes

Things that could be improved:
- Nightly doesn't handle line breaks in the message. The displayed message is not very nice.
- Phantom does handle line breaks, but there's no automatic wrapping, so the overflow is just hidden and we need to expand the prompt

## Todos (upcoming PRs)
- [ ] Decide on a good format for displaying transaction info in human-readable but also concise way.
- [ ] Write example on how to use derived wallets for cross-chain needs

## Demos

Phantom (Solana):
- Fully functional
- Message is formatted as expected
- Does not emit events when switching active account

https://github.com/user-attachments/assets/17a885fa-4866-4141-9cca-738c61080a1f

Nightly (Solana):
- `solana:signIn` not implemented, can't sign with derived wallet
- Emits events when the active account is switched
- Doesn't emit when account is disconnected from the wallet

https://github.com/user-attachments/assets/49da51a1-f5e3-4135-90c4-da8a5fc0ed35

Metamask (Ethereum):
- Fully functional
- Emits events when connected accounts are changed (including full disconnect)
- No native concept of "active account", we consider the first connected account as active

https://github.com/user-attachments/assets/58b99de1-a7e2-4ac5-950a-b8aa65ebd16e

Nightly (Ethereum):
- Fully functional
- Emits events when the active account is switched
- Doesn't emit when account is disconnected from the wallet
- Newlines are skipped in the message, formatting is not great

https://github.com/user-attachments/assets/9f2bdc62-d1c0-4da2-938c-b73d77576908

### DIFF
--- a/apps/nextjs-example/src/components/WalletProvider.tsx
+++ b/apps/nextjs-example/src/components/WalletProvider.tsx
@@ -1,11 +1,27 @@
 "use client";
 
 import { AptosWalletAdapterProvider } from "@aptos-labs/wallet-adapter-react";
+import { setupAutomaticEthereumWalletDerivation } from '@aptos-labs/derived-wallet-ethereum';
+import { setupAutomaticSolanaWalletDerivation } from '@aptos-labs/derived-wallet-solana';
 import { PropsWithChildren } from "react";
 import { Network } from "@aptos-labs/ts-sdk";
 import { useClaimSecretKey } from "@/hooks/useClaimSecretKey";
 import { useAutoConnect } from "./AutoConnectProvider";
 import { useToast } from "./ui/use-toast";
+
+const searchParams = typeof window !== "undefined" ? new URL(window.location.href).searchParams : undefined
+const deriveWalletsFrom = searchParams?.get('deriveWalletsFrom')?.split(',');
+if (deriveWalletsFrom?.includes('ethereum')) {
+  setupAutomaticEthereumWalletDerivation({ defaultNetwork: Network.TESTNET });
+}
+if (deriveWalletsFrom?.includes('solana')) {
+  setupAutomaticSolanaWalletDerivation({ defaultNetwork: Network.TESTNET });
+}
+
+let dappImageURI: string | undefined;
+if (typeof window !== 'undefined') {
+  dappImageURI = `${window.location.origin}${window.location.pathname}favicon.ico`;
+}
 
 export const WalletProvider = ({ children }: PropsWithChildren) => {
   const { autoConnect } = useAutoConnect();
@@ -26,6 +42,7 @@ export const WalletProvider = ({ children }: PropsWithChildren) => {
         aptosConnect: {
           claimSecretKey,
           dappId: "57fa42a9-29c6-4f1e-939c-4eefa36d9ff5",
+          dappImageURI,
         },
         mizuwallet: {
           manifestURL:
--- a/packages/derived-wallet-ethereum/src/EIP1193DerivedWallet.ts
+++ b/packages/derived-wallet-ethereum/src/EIP1193DerivedWallet.ts
@@ -1,0 +1,337 @@
+import {
+  encodeStructuredMessage,
+  isNullCallback,
+  makeUserApproval,
+  makeUserRejection,
+  mapUserResponse,
+} from '@aptos-labs/derived-wallet-base';
+import {
+  AccountAuthenticator,
+  AccountAuthenticatorAbstraction,
+  AnyRawTransaction,
+  Network,
+  NetworkToChainId,
+  NetworkToNodeAPI,
+} from '@aptos-labs/ts-sdk';
+import {
+  AccountInfo,
+  APTOS_CHAINS,
+  AptosChangeNetworkOutput,
+  AptosConnectOutput,
+  AptosFeatures,
+  AptosSignMessageInput,
+  AptosSignMessageOutput,
+  AptosWallet,
+  NetworkInfo,
+  UserResponse,
+  UserResponseStatus,
+  WalletIcon,
+} from "@aptos-labs/wallet-standard";
+import type { EIP6963ProviderDetail } from 'mipd';
+import {
+  Address as EthereumAddress,
+  createWalletClient,
+  custom,
+  CustomTransport,
+  EIP1193Provider,
+  UserRejectedRequestError,
+  WalletClient,
+} from 'viem'
+import {
+  createSiweMessageFromAptosStructuredMessage,
+  createSiweMessageFromAptosTransaction,
+} from './createSiweMessageFromAptos';
+import { EIP1193DerivedPublicKey } from './EIP1193DerivedPublicKey';
+import { EIP1193DerivedSignature } from './EIP1193DerivedSignature';
+
+const defaultAuthenticationFunction = '0x7::eip1193::authenticate';
+
+/**
+ * Adapt EIP1193 response into a UserResponse.
+ * `UserRejectedRequestError` will be converted into a rejection.
+ */
+async function wrapEIP1193UserResponse<TResponse>(promise: Promise<TResponse>): Promise<UserResponse<TResponse>> {
+  try {
+    const response = await promise;
+    return makeUserApproval(response);
+  } catch (err) {
+    if (err instanceof UserRejectedRequestError) {
+      return makeUserRejection();
+    }
+    throw err;
+  }
+}
+
+export interface EIP1193DerivedWalletOptions {
+  authenticationFunction?: string;
+  defaultNetwork?: Network;
+}
+
+export class EIP1193DerivedWallet implements AptosWallet {
+  readonly eip1193Provider: EIP1193Provider;
+  readonly eip1193Wallet: WalletClient<CustomTransport>;
+  readonly domain: string;
+  readonly authenticationFunction: string;
+  defaultNetwork: Network;
+
+  readonly version = "1.0.0";
+  readonly name: string;
+  readonly icon: WalletIcon;
+  readonly url: string;
+  readonly accounts = [];
+  readonly chains = APTOS_CHAINS;
+
+  constructor(providerDetail: EIP6963ProviderDetail, options: EIP1193DerivedWalletOptions = {}) {
+    const { info, provider } = providerDetail;
+    const {
+      authenticationFunction = defaultAuthenticationFunction,
+      defaultNetwork = Network.MAINNET,
+    } = options;
+
+    this.eip1193Provider = provider;
+    this.eip1193Wallet = createWalletClient({
+      transport: custom(this.eip1193Provider),
+    })
+
+    this.domain = window.location.origin;
+    this.authenticationFunction = authenticationFunction;
+    this.defaultNetwork = defaultNetwork;
+    this.name = `${info.name} (Ethereum)`;
+    // Phantom's icon is wrapped with new lines :shrug:
+    this.icon = info.icon.trim() as WalletIcon;
+    this.url = info.rdns;
+  }
+
+  readonly features: AptosFeatures = {
+    'aptos:connect': {
+      version: '1.0.0',
+      connect: () => this.connect(),
+    },
+    'aptos:disconnect': {
+      version: '1.0.0',
+      disconnect: () => this.disconnect(),
+    },
+    'aptos:account': {
+      version: '1.0.0',
+      account: () => this.getActiveAccount(),
+    },
+    'aptos:onAccountChange': {
+      version: '1.0.0',
+      onAccountChange: async (callback) => this.onActiveAccountChange(callback),
+    },
+    'aptos:network': {
+      version: '1.0.0',
+      network: () => this.getActiveNetwork(),
+    },
+    'aptos:changeNetwork': {
+      version: '1.0.0',
+      changeNetwork: (newNetwork) => this.changeNetwork(newNetwork),
+    },
+    'aptos:onNetworkChange': {
+      version: '1.0.0',
+      onNetworkChange: async (callback) => this.onActiveNetworkChange(callback),
+    },
+    "aptos:signMessage": {
+      version: '1.0.0',
+      signMessage: (args) => this.signMessage(args),
+    },
+    "aptos:signTransaction": {
+      version: '1.0.0',
+      signTransaction: (...args) => this.signTransaction(...args),
+    },
+  }
+
+  private derivePublicKey(ethereumAddress: EthereumAddress) {
+    return new EIP1193DerivedPublicKey({
+      domain: this.domain,
+      ethereumAddress,
+      authenticationFunction: this.authenticationFunction,
+    });
+  }
+
+  // region Connection
+
+  async connect(): Promise<UserResponse<AptosConnectOutput>> {
+    const response = await wrapEIP1193UserResponse(this.eip1193Wallet.requestAddresses());
+
+    return mapUserResponse(response, ([ethereumAddress]) => {
+      const publicKey = this.derivePublicKey(ethereumAddress);
+      const aptosAddress = publicKey.authKey().derivedAddress();
+      return new AccountInfo({ publicKey, address: aptosAddress });
+    });
+  }
+
+  async disconnect() {
+    // TODO: Eip1193 doesn't provide a "disconnect" method, so we have to keep track locally
+  }
+
+  // endregion
+
+  // region Accounts
+
+  private async getActiveAccountAddress(): Promise<EthereumAddress> {
+    const [activeAddress] = await this.eip1193Wallet.getAddresses();
+    if (!activeAddress) {
+      throw new Error('Account not connected');
+    }
+    return activeAddress;
+  }
+
+  async getActiveAccount() {
+    const ethereumAddress = await this.getActiveAccountAddress();
+    const publicKey = this.derivePublicKey(ethereumAddress);
+    const aptosAddress = publicKey.authKey().derivedAddress();
+    return new AccountInfo({ publicKey, address: aptosAddress });
+  }
+
+  private onAccountsChangedListeners: ((newAccounts: EthereumAddress[]) => void)[] = []
+
+  onActiveAccountChange(callback: (newAccount: AccountInfo) => void) {
+    if (isNullCallback(callback)) {
+      for (const listener of this.onAccountsChangedListeners) {
+        this.eip1193Provider.removeListener('accountsChanged', listener);
+      }
+      this.onAccountsChangedListeners = [];
+    } else {
+      const listener = ([ethereumAddress]: EthereumAddress[]) => {
+        if (!ethereumAddress) {
+          callback(undefined as any as AccountInfo);
+          return;
+        }
+        const publicKey = this.derivePublicKey(ethereumAddress);
+        const aptosAddress = publicKey.authKey().derivedAddress();
+        const account = new AccountInfo({ publicKey, address: aptosAddress });
+        callback(account);
+      };
+      this.onAccountsChangedListeners.push(listener);
+      this.eip1193Provider.on('accountsChanged', listener);
+    }
+  }
+
+  // endregion
+
+  // region Networks
+
+  private onActiveNetworkChangeListeners: ((newNetwork: NetworkInfo) => void)[] = [];
+
+  async getActiveNetwork(): Promise<NetworkInfo> {
+    const chainId = NetworkToChainId[this.defaultNetwork];
+    const url = NetworkToNodeAPI[this.defaultNetwork];
+    return {
+      name: this.defaultNetwork,
+      chainId,
+      url,
+    };
+  }
+
+  async changeNetwork(newNetwork: NetworkInfo): Promise<UserResponse<AptosChangeNetworkOutput>> {
+    const { name, chainId, url } = newNetwork;
+    if (name === Network.CUSTOM) {
+      throw new Error('Custom network not currently supported');
+    }
+    this.defaultNetwork = name;
+    for (const listener of this.onActiveNetworkChangeListeners) {
+      listener({
+        name,
+        chainId: chainId ?? NetworkToChainId[name],
+        url: url ?? NetworkToNodeAPI[name],
+      });
+    }
+    return {
+      status: UserResponseStatus.APPROVED,
+      args: { success: true },
+    };
+  }
+
+  onActiveNetworkChange(callback: (newNetwork: NetworkInfo) => void) {
+    if (isNullCallback(callback)) {
+      this.onActiveNetworkChangeListeners = [];
+    } else {
+      this.onActiveNetworkChangeListeners.push(callback);
+    }
+  }
+
+  // endregion
+
+  // region Signatures
+
+  async signMessage({
+    message,
+    nonce,
+    ...flags
+  }: AptosSignMessageInput): Promise<UserResponse<AptosSignMessageOutput>> {
+    const ethereumAddress = await this.getActiveAccountAddress();
+    const publicKey = this.derivePublicKey(ethereumAddress);
+
+    const aptosAddress = flags.address ? publicKey.authKey().derivedAddress() : undefined;
+    const application = flags.application ? this.domain : undefined;
+    const chainId = flags.chainId ? NetworkToChainId[this.defaultNetwork] : undefined;
+    const structuredMessage = {
+      address: aptosAddress?.toString(),
+      application,
+      chainId,
+      message,
+      nonce,
+    };
+
+    // We need to provide `issuedAt` externally so that we can match it with the signature
+    const issuedAt = new Date();
+    const siweMessage = createSiweMessageFromAptosStructuredMessage({
+      ethereumAddress,
+      structuredMessage,
+      issuedAt,
+    });
+
+    const response = await wrapEIP1193UserResponse(this.eip1193Wallet.signMessage({
+      account: ethereumAddress,
+      message: siweMessage,
+    }));
+
+    return mapUserResponse(response, (siweSignature) => {
+      const signature = new EIP1193DerivedSignature(siweSignature, issuedAt);
+      const fullMessageBytes = encodeStructuredMessage(structuredMessage);
+      const fullMessage = new TextDecoder().decode(fullMessageBytes);
+      return {
+        prefix: 'APTOS',
+        fullMessage,
+        message,
+        nonce,
+        signature,
+      };
+    });
+  }
+
+  async signTransaction(rawTransaction: AnyRawTransaction, asFeePayer?: boolean): Promise<UserResponse<AccountAuthenticator>> {
+    const ethereumAddress = await this.getActiveAccountAddress();
+    const publicKey = this.derivePublicKey(ethereumAddress);
+    const aptosAddress = publicKey.authKey().derivedAddress();
+
+    // We need to provide `issuedAt` externally so that we can match it with the signature
+    const issuedAt = new Date();
+    const siweMessage = createSiweMessageFromAptosTransaction({
+      aptosAddress,
+      ethereumAddress,
+      rawTransaction,
+      issuedAt,
+    });
+
+    const response = await wrapEIP1193UserResponse(this.eip1193Wallet.signMessage({
+      account: ethereumAddress,
+      message: siweMessage,
+    }));
+
+    return mapUserResponse(response, (siweSignature) => {
+      const utf8EncodedSiweMessage = new TextEncoder().encode(siweMessage);
+      const signature = new EIP1193DerivedSignature(siweSignature, issuedAt);
+      const authenticator: AccountAuthenticator = new AccountAuthenticatorAbstraction(
+        this.authenticationFunction,
+        // Not sure what the expected value is here
+        utf8EncodedSiweMessage,
+        signature.bcsToBytes(),
+      );
+      return authenticator;
+    });
+  }
+
+  // endregion
+}
--- a/packages/derived-wallet-solana/src/SolanaDerivedWallet.ts
+++ b/packages/derived-wallet-solana/src/SolanaDerivedWallet.ts
@@ -1,0 +1,322 @@
+import {
+  accountInfoFromPublicKey,
+  encodeStructuredMessage,
+  isNullCallback,
+  makeUserApproval,
+  makeUserRejection,
+  mapUserResponse,
+} from '@aptos-labs/derived-wallet-base';
+import {
+  AccountAuthenticator,
+  AccountAuthenticatorAbstraction,
+  AnyRawTransaction,
+  Ed25519Signature,
+  Network,
+  NetworkToChainId,
+  NetworkToNodeAPI,
+} from '@aptos-labs/ts-sdk';
+import {
+  AccountInfo,
+  APTOS_CHAINS,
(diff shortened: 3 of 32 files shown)


packages/from-solana/src/SolanaDomainWallet.ts
```diff
@@ -0,0 +1,255 @@
...
+      version: '1.0.0',
+      disconnect: () => this.disconnect(),
+    },
+    'aptos:account': {
+      version: '1.0.0',
+      account: () => this.getActiveAccount(),
+    },
+    'aptos:onAccountChange': {
+      version: '1.0.0',
+      // Theoretically supported, but not sure how
```
0xmaayan: can refer to https://github.com/aptos-labs/aptos-wallet-adapter/pull/500/files#diff-907d94e30b45520090d144d8500765e27e7a4bb68a6a5dbcb865ea8ed7fe5d94R147-R164

packages/from-solana/src/SolanaDomainWallet.ts
```diff
@@ -0,0 +1,255 @@
...
+    if (!this.solanaWallet.publicKey) {
+      throw new Error('Account not connected');
+    }
+    return {
+      domain: this.domain,
+      solanaPublicKey: this.solanaWallet.publicKey,
+    };
+  }
+
+  async connect(): Promise<UserResponse<AptosConnectOutput>> {
```
0xmaayan: this should return the origin wallet account info

packages/from-solana/src/SolanaDomainWallet.ts
```diff
@@ -0,0 +1,255 @@
...
+    // console.log(this.solanaWallet.name, 'initial value', this.solanaWallet.wallet.accounts);
+    // eventsFeature.on('change', (args) => {
+    //   console.log('changed', args);
+    // });
+    // this.solanaWallet.on('connect', (account) => {
+    //   console.log(this.solanaWallet.name, 'connected', account);
+    // });
+  }
+
+  readonly features: AptosFeatures = {
```
0xmaayan: we are missing `signIn` feature https://github.com/aptos-labs/aptos-wallet-adapter/pull/500/files#diff-907d94e30b45520090d144d8500765e27e7a4bb68a6a5dbcb865ea8ed7fe5d94R181

0xmaayan (commented):
Nice work!

I like how each wallet has (${sourceChain}) attached to it.

I think we should separate the x-chain wallets and the existing aptos-only wallets provider and/or UI. This dapp specific DAA implementation is to answer a specific dapp use case and not to be used by all dapps.

Also, we should treat the wallet as origin-chain first, which means when asking for the `account()` or `connect()` we should get the origin chain wallet info, and expose a separate feature/API to get the `aptosAccount()` (which is the derived DAA)

I do see how this implementation can work with the [other PR implementation ](https://github.com/aptos-labs/aptos-wallet-adapter/pull/500)

packages/from-solana/src/SolanaDomainWallet.ts
hardsetting: mm you mean like Solana's?
This is the implementation of `aptos:connect` though

apps/nextjs-example/src/components/WalletProvider.tsx
```diff
@@ -1,12 +1,20 @@
...
 
 import { AptosWalletAdapterProvider } from "@aptos-labs/wallet-adapter-react";
+import { setupSolanaDomainAutomaticConversion } from '@aptos-labs/wallet-from-solana';
 import { PropsWithChildren } from "react";
 import { Network } from "@aptos-labs/ts-sdk";
 import { useClaimSecretKey } from "@/hooks/useClaimSecretKey";
 import { useAutoConnect } from "./AutoConnectProvider";
 import { useToast } from "./ui/use-toast";
 
+setupSolanaDomainAutomaticConversion({ defaultNetwork: Network.TESTNET });
```
0xmaayan: thoughts about renaming? maybe `registerSolanaDomainWallets({...})`

packages/from-solana/src/SolanaDomainWallet.ts
```diff
@@ -0,0 +1,255 @@
...
+  constructor(solanaWallet: SolanaWalletAdapter, options: SolanaDomainWalletOptions = {}) {
+    const { domain, defaultNetwork = Network.MAINNET } = options;
+    this.solanaWallet = solanaWallet;
+    this.domain = domain ?? window.origin;
+    this.defaultNetwork = defaultNetwork;
+    this.name = `${solanaWallet.name} (Solana)`;
+    this.icon = solanaWallet.icon;
+    this.url = solanaWallet.url;
+
+    // const eventsFeature = this.solanaWallet.wallet.features['standard:events'];
```
0xmaayan: remove?

packages/from-solana/src/SolanaDomainWallet.ts
0xmaayan: need to register to events in the connect method https://github.com/aptos-labs/aptos-wallet-adapter/pull/500/files#diff-2dbb4ee347e6ed68f359340210173faf3a1a33599bd885aea23bc68343023377R113

packages/from-solana/src/SolanaDomainWallet.ts
```diff
@@ -0,0 +1,255 @@
...
+                    }: AptosSignMessageInput): Promise<UserResponse<AptosSignMessageOutput>> {
+    if (!this.solanaWallet.signMessage) {
+      throw new Error('solana:signMessage not available');
+    }
+
+    const solanaIdentity = this.activeSolanaIdentity;
+    const publicKey = new SolanaDomainPublicKey(solanaIdentity, authenticationFunction);
+
+    // {
+    //   const messageBytes = Hex.fromHexInput('0xdeadbeef').toUint8Array();
```
0xmaayan: remove?

packages/from-solana/src/SolanaDomainWallet.ts
```diff
@@ -0,0 +1,255 @@
...
+    );
+
+    let signatureBytes: Uint8Array;
+    try {
+      signatureBytes = await this.solanaWallet.signMessage(solanaSigningMessage);
+    } catch (err) {
+      if (err instanceof WalletError) {
+        return { status: UserResponseStatus.REJECTED };
+      }
+      throw err;
```
0xmaayan: With how the wallet core is built, that would break the dapp, we need to
```
throw new Error(err).message
```

packages/from-solana/src/SolanaDomainWallet.ts
```diff
@@ -0,0 +1,255 @@
...
+        message,
+        nonce,
+        signature,
+      },
+    };
+  }
+
+  async signTransaction(rawTransaction: AnyRawTransaction, asFeePayer?: boolean): Promise<UserResponse<AccountAuthenticator>> {
+    if (!this.solanaWallet.signMessage) {
+      throw new Error('solana:signMessage not available');
```
0xmaayan:
```suggestion
      throw new Error('solana:signMessage not available').message;
```

packages/from-solana/src/SolanaDomainWallet.ts
```diff
@@ -0,0 +1,255 @@
...
+    const solanaSigningMessageDigest = hashValues([solanaSigningMessage]);
+
+    let signatureBytes: Uint8Array;
+    try {
+      signatureBytes = await this.solanaWallet.signMessage(solanaSigningMessage);
+    } catch (err) {
+      if (err instanceof WalletError) {
+        return { status: UserResponseStatus.REJECTED };
+      }
+      throw err;
```
0xmaayan: same
```suggestion
      throw new Error(`Failed to sign transaction ${err}`).message;
```

packages/from-solana/src/crypto/SolanaDomainPublicKey.ts
```diff
@@ -0,0 +1,58 @@
...
+  hashValues,
+  Hex,
+  Serializer,
+  VerifySignatureArgs,
+} from '@aptos-labs/ts-sdk';
+import { computeDomainAuthenticationKey } from './abstraction';
+import { makeSolanaSigningMessage } from './makeSolanaSigningMessage';
+import { SolanaDomainIdentity } from './SolanaDomainIdentity';
+
+// Should extend AbstractPublicKey, but it's currently not exported
```
0xmaayan: I got you https://github.com/aptos-labs/aptos-ts-sdk/pull/661

packages/from-solana/src/crypto/abstraction.ts
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Compute the account address of the DAA
+ * The DAA account address is computed by hashing the function info and the account identity
+ * and appending the domain separator (5)
+ *
+ * @param functionInfo - The authentication function
+ * @param accountIdentifier - The account identity
+ * @returns The account address
+ */
+export function computeDomainAuthenticationKey(functionInfo: string, accountIdentifier: Uint8Array) {
```
0xmaayan: This whole logic will soon be in the aptos ts-sdk https://github.com/aptos-labs/aptos-ts-sdk/pull/648/files#diff-4e9f5a9b3686421e45ff1d01238ada042f65b998021ae9235a3329bfa019c695R68

packages/from-solana/src/SolanaDomainWallet.ts
```diff
@@ -0,0 +1,255 @@
...
+    const chainId = NetworkToChainId[this.defaultNetwork];
+    const url = NetworkToNodeAPI[this.defaultNetwork];
+    return {
+      name: this.defaultNetwork,
+      chainId,
+      url,
+    };
+  }
+
+  async signMessage({
```
0xmaayan: This whole flow is more relevant to the future `signIn()` implementation. `signMessage` is pretty simple and dump compare to the other and is just signing an encoded utf8 message
https://docs.phantom.com/solana/signing-a-message

packages/from-solana/src/SolanaDomainWallet.ts
```diff
@@ -0,0 +1,255 @@
...
+    //   const signature = new Ed25519Signature(signatureBytes);
+    //   pubKey.verifySignature({ message: messageBytes, signature });
+    //   console.log('WORKS');
+    // }
+
+    const address = flags.address ? publicKey.authKey().derivedAddress().toString() : undefined;
+    const application = flags.application ? this.domain : undefined;
+    const chainId = NetworkToChainId[this.defaultNetwork];
+    const structuredMessage = { address, application, chainId, message, nonce };
+    const encodedMessage = encodeStructuredMessage(structuredMessage);
```
0xmaayan: once this will be part of the `signIn()` method, the format is being done automatically by the wallet, so just need to provide the fields
https://github.com/phantom/sign-in-with-solana?tab=readme-ov-file#sign-in-input-fields

And can check the implementation on the other PR 
https://github.com/aptos-labs/aptos-wallet-adapter/pull/500/files#diff-907d94e30b45520090d144d8500765e27e7a4bb68a6a5dbcb865ea8ed7fe5d94R193
and how it looks like in the live demo https://aptos-labs.github.io/aptos-wallet-adapter/nextjs-example-testing/swap

packages/from-solana/src/crypto/SolanaDomainPublicKey.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  authKey(): AuthenticationKey {
+    return this._authKey;
+  }
+
+  verifySignature({ message, signature }: VerifySignatureArgs): boolean {
+    const messageBytes = Hex.fromHexInput(message).toUint8Array();
+    const solanaSigningMessage = makeSolanaSigningMessage(this.identity, messageBytes);
+
+    // Solana uses Ed25519 just like Aptos
```
0xmaayan: if solana uses the same ED25519 can't we just extend/use the Ed25519PublicKey class from the ts-sdk?

0xmaayan (commented):
Looks great! Some notes
- We need the `signIn()` implementation (can refer to the other [PR](https://github.com/aptos-labs/aptos-wallet-adapter/pull/500/files#diff-907d94e30b45520090d144d8500765e27e7a4bb68a6a5dbcb865ea8ed7fe5d94R174)). I imagine the flow to be something like (can make adjustments as needed):
  - Dapp builds the transaction (aptos.transaction.build.simple({...}) and sends it to the wallet
  - Adapter generates the transaction hash
  - Adapter sends the minimum required fields (nonce, txn hash) to the wallet
  - The wallet will construct a message to sign with the provided fields and the account address (would be great to display the function info (i.e `address::module::function` in a human readable format on the wallet ui)). 
  - User signs the message and the adapter verifies the signature
  - Adapter sends the signature back to the dapp to submit the transaction with the signature to chain
- Lets land this change on `main` branch so we can keep executing, the new interface PR is good but need to think on the migration path

packages/from-solana/src/SolanaDomainWallet.ts
0xmaayan: We need a feature or an exposed API to return the current origin wallet account info (address, public key, etc)

hardsetting: let's add this in another PR

packages/from-solana/src/SolanaDomainWallet.ts
hardsetting: I had an initial implementation for Solana, but paused this as it's not essential (SIWA is not a required feature).
We can add support later

apps/nextjs-example/src/components/WalletProvider.tsx
hardsetting: I called this `setupAutomaticSolanaWalletDerivation`:
- removed "domain" in favor of "derivation" to match the new AIP's name
- we should still keep "setupAutomatic..." as this suggests we're setting up listeners, and that a cleanup step also exists (the cleanup callback we return)

We could also call it `setupAutomaticWalletDerivationFromSolana` to hint more strongly that we're deriving an Aptos wallet from a Solana wallet.

packages/from-solana/src/SolanaDomainWallet.ts
hardsetting: I think the wallet core is catching errors and rethrowing them like you mentioned.
Luckily that practice can be isolated to the wallet core package only :D

packages/from-solana/src/SolanaDomainWallet.ts
hardsetting: I changed the implementation to use Solana's and Ethereum's `signIn` features for message verification

packages/from-solana/src/crypto/SolanaDomainPublicKey.ts
hardsetting: In which context? Like extending `SolanaDerivedPublicKey` from `Ed25519PublicKey`?
I'd rather go with composition rather than subclassing in this case, it wouldn't add much value extending the class.

apps/nextjs-example/src/components/WalletProvider.tsx
```diff
@@ -1,12 +1,28 @@
...
 import { AptosWalletAdapterProvider } from "@aptos-labs/wallet-adapter-react";
+import { setupAutomaticEthereumWalletDerivation } from '@aptos-labs/derived-wallet-ethereum';
+import { setupAutomaticSolanaWalletDerivation } from '@aptos-labs/derived-wallet-solana';
 import { PropsWithChildren } from "react";
 import { Network } from "@aptos-labs/ts-sdk";
 import { useClaimSecretKey } from "@/hooks/useClaimSecretKey";
 import { useAutoConnect } from "./AutoConnectProvider";
 import { useToast } from "./ui/use-toast";
 
+const searchParams = typeof window !== "undefined" ? new URL(window.location.href).searchParams : undefined
```
0xmaayan, line 12: nice. 

I still think we should not include it in this page, but in a separate demo page - x-chain wallets have different implementations that not relevant to the only aptos wallets flow.

Let's just merge the wallet packages, and then I can use it as part of my other PR that introduces the cross chain sdk and the demo dapp

packages/derived-wallet-solana/src/SolanaDerivedWallet.ts
```diff
@@ -0,0 +1,322 @@
...
+    } else {
+      this.onActiveNetworkChangeListeners.add(callback);
+    }
+  }
+
+  // endregion
+
+  // region Signatures
+
+  async signMessage({
```
0xmaayan, line 234: thoughts on simply calling it `signIn()`, and have a `signMessage()` throw "not yet implemented"

packages/derived-wallet-solana/src/SolanaDerivedWallet.ts
```diff
@@ -0,0 +1,322 @@
...
+  }: AptosSignMessageInput): Promise<UserResponse<AptosSignMessageOutput>> {
+    if (!this.solanaWallet.signIn) {
+      throw new Error('solana:signIn not available');
+    }
+    const aptosPublicKey = this.getActivePublicKey();
+    const aptosAddress = flags.address ? aptosPublicKey.authKey().derivedAddress() : undefined;
+    const application = flags.application ? this.domain : undefined;
+    const chainId = flags.chainId ? NetworkToChainId[this.defaultNetwork] : undefined;
+    const structuredMessage = {
+      address: aptosAddress?.toString(),
```
0xmaayan, line 247: is that the DAA address? we should show the actual wallet (solana) address as this is the wallet that signs the message

packages/derived-wallet-solana/src/SolanaDerivedWallet.ts
```diff
@@ -0,0 +1,322 @@
...
+        signature,
+      };
+    });
+  }
+
+  async signTransaction(
+    rawTransaction: AnyRawTransaction,
+    _asFeePayer?: boolean,
+  ): Promise<UserResponse<AccountAuthenticator>> {
+    if (!this.solanaWallet.signIn) {
```
0xmaayan, line 287: we should be able to sign transaction even if the wallet has not implemented SIWS, also the actual signer should be the original solana wallet?

I suggest keeping this implementation with the basic Solana sign transaction standard

packages/derived-wallet-ethereum/src/EIP1193DerivedWallet.ts
```diff
@@ -0,0 +1,337 @@
...
+
+    return mapUserResponse(response, ([ethereumAddress]) => {
+      const publicKey = this.derivePublicKey(ethereumAddress);
+      const aptosAddress = publicKey.authKey().derivedAddress();
+      return new AccountInfo({ publicKey, address: aptosAddress });
+    });
+  }
+
+  async disconnect() {
+    // TODO: Eip1193 doesn't provide a "disconnect" method, so we have to keep track locally
```
0xmaayan, line 165: ? https://github.com/aptos-labs/aptos-wallet-adapter/pull/500/files#diff-1e5ca5da2435cf31e8d3db861165c0d573e2d6a7a20db76d92394c790174b1a7R57-R67

0xmaayan (commented):
Looks great! some comments:
- Can we remove the implementation on the existing nextjs demo? I'll create a new demo specially for that that includes more relevant x-chain functionalities and excludes other functionalities - would be easier to test and make changes while we are working on the DEX product
- `connect()` method should return the origin wallet info, so the dapp can display it on the UI, right now it seems it returns the DAA/aptos account. Or we need a wallet API to `getOriginalWallet()`
- Similarly, we need an API to `getOriginalNetwork()` (i.e the actual evm network the wallet is connected to, not applicable to Solana)
- Looks like we use `signInWith*()` implementation for `signMessage()` and `signTransaction()`. I prefer having both functions implementation as it should be, and add a `signIn()` feature specifically for `signInWith*()`. 
- Other operations, like [CCTP transfer](https://github.com/aptos-labs/aptos-wallet-adapter/pull/500/files#diff-49a2918581a5c1570378baaaa48574506401b3187797a3b2cdbd36f65eb224beR18-R35), rely on some of the EVM wallet original implementations,  can we support [sendTransaction()](https://github.com/aptos-labs/aptos-wallet-adapter/pull/500/files#diff-46880f411a46bc01096ba76b4938e04b2ce3fd4d73b35834471cd9c94262e66bR217) also?

apps/nextjs-example/src/components/WalletProvider.tsx
hardsetting: I'm fine removing this, but also how would be we testing the basic wallet functionalities for derived wallets?
e.g. signMessage, signAndSubmitTransaction, signTransaction with options..
as opposed to cross-chain specific features

What's the harm of keeping? We need an explicit query param to enable automatic derivation

packages/derived-wallet-solana/src/SolanaDerivedWallet.ts
hardsetting: uhm I don't follow, why would we do that?

hardsetting: We're using `solana:signIn` to sign Aptos messages (whether that's a structured message, or a raw transaction).
We could also implement `aptos:signIn`, which would also call `solana:signIn` in the implementation.

packages/derived-wallet-solana/src/SolanaDerivedWallet.ts
hardsetting: Yes that's the DAA address (Aptos).
The structured message is an Aptos object, so we need the Aptos address here

The solana address is needed for example when creating the SIWS message, in the Solana context

packages/derived-wallet-solana/src/SolanaDerivedWallet.ts
hardsetting: Yes we could, we should probably check with @igor-aptos and @alinush though.

I think they wouldn't like this as it's allowing unverified signatures.
Basically if you're using Nightly with Solana, you're susceptible to phishing attacks as there's nothing preventing `evil-dapp.com` from pretending to be `good-dapp.com`. This plus the fact we don't have simulations is a very scary combo.

hardsetting: To be fair though, I think Nightly allows signing a SIWS challenge as `signMessage` which is bad.

`evil-dapp.com` could fork the wallet adapter and use `signMessage` only when Nightly is detected.
The user would happily sign a `totally_legit_module::definitely_not_draining_your_funds` entry function and then get drained because there's no simulation.

We should actually encourage people to not use Nightly with Solana if that's the case 😬 
Maybe I can do a quick test before making this conclusion.

hardsetting: >  the actual signer should be the original solana wallet?

uhm could you elaborate more on why?
The transaction needs to be signed by an Aptos account, in this case the account derived from the solana account

packages/derived-wallet-ethereum/src/EIP1193DerivedWallet.ts
hardsetting: The implementation you linked me is setting up an event listener.
How would that trigger a disconnect?

apps/nextjs-example/src/components/WalletProvider.tsx
0xmaayan: I'd like to put CCTP transfer on the demo app, and dont want to start dealing with if(searchParam){....}. Also, devs usually refer to this demo page code base to implement different functionalities, having both x-cahin and only aptos wallets on the same dapp may confuse them.

We can add any feature we want to test to the demo app. This is mostly for while we are developing the DEX product as we can expect changes and bugs and dont want to affect the existing products/flows

packages/derived-wallet-ethereum/src/EIP1193DerivedWallet.ts
hardsetting: "I want the ethereum wallet to disconnect, so I wait for it to tell me when it disconnects and when that happens I do nothing on success, or rethrow on error"

packages/derived-wallet-solana/src/SolanaDerivedWallet.ts
0xmaayan: What I mean is, on the wallet UI, where we display the message, we should display the Solana wallet address and not the DAA one as the "wants you to sign in with your {wallet} account ____"

hardsetting:
> Can we remove the implementation on the existing nextjs demo? I'll create a new demo specially for that that includes more relevant x-chain functionalities and excludes other functionalities - would be easier to test and make changes while we are working on the DEX product

Responded in the comment

> connect() method should return the origin wallet info, so the dapp can display it on the UI, right now it seems it returns the DAA/aptos account. Or we need a wallet API to getOriginalWallet()

`aptos:connect` should return an Aptos account.

Yes we shoud provide access to the source wallet.
If the user derived the wallet manually, they always have access to the source e.g.
```typescript
const solanaWallet = new SolanaWalletAdapter(...);
const derivedAptosWallet = new SolanaDerivedWallet(solanaWallet);

solanaWallet.signTransaction(...);
derivedAptosWallet.signTransaction(...);
```
This would happen for example if a dapp implements wallet discovery for Solana and has access to the solana wallet first.

On the other hand, If the user is using the Aptos wallet discovery mechanism then:
```typescript
const activeWallet = useActiveWallet();
if (activeWallet.wallet instanceof SolanaDerivedWallet) {
  const aptosWallet = activeWallet;
  const solanaWallet = activeWallet.wallet;

  solanaWallet.signTransaction(...);
  aptosWallet.signTransaction(...);
}
```

> Similarly, we need an API to getOriginalNetwork() (i.e the actual evm network the wallet is connected to, not applicable to Solana)

We're never actually signing Solana transactions on the Solana blockchain **through the derived Aptos wallet**.
If we're signing Solana transactions (e.g. for CCTP), that can happen through the solana wallet adapter, and we have access to all Solana features there

> Looks like we use signInWith*() implementation for signMessage() and signTransaction(). I prefer having both functions implementation as it should be, and add a signIn() feature specifically for signInWith*().

I think we might not be on the same page on why we're using SIWS and SIWE at the moment.
Let's say we need to sign a transaction on the Aptos blockchain using a DAA account derived from Solana.
We need to build an Aptos `RawTransaction` as usual, and then sign in.
But how do we sign it if the DAA account has no private key? What would the signature even look like?
The AIP proposal says that we can "transform" an Aptos signing message (in this case the transaction) into a Solana signing message, and then sign it with the Solana account.
But Solana allows for arbitrary signatures, which is bad. This means we have no protection over who the dapp claim to be.
On the other hand the SIWS standard comes with a `domain` and `URI` fields and the wallet will complain if the dapp's domain doesn't match.
That's why we're choosing to use SIWS to sign messages even though it's a "sign in" feature.

> Other operations, like [CCTP transfer](https://github.com/aptos-labs/aptos-wallet-adapter/pull/500/files#diff-49a2918581a5c1570378baaaa48574506401b3187797a3b2cdbd36f65eb224beR18-R35), rely on some of the EVM wallet original implementations, can we support [sendTransaction()](https://github.com/aptos-labs/aptos-wallet-adapter/pull/500/files#diff-46880f411a46bc01096ba76b4938e04b2ce3fd4d73b35834471cd9c94262e66bR217) also?

As mentioned above, we should just obtain the source chain adapter and use that

packages/derived-wallet-solana/src/SolanaDerivedWallet.ts
0xmaayan: > uhm could you elaborate more on why?
> The transaction needs to be signed by an Aptos account, in this case the account derived from the solana account

The message needs to be signed by the original wallet private key, i.e solana or evm

apps/nextjs-example/src/components/WalletProvider.tsx
hardsetting: I agree we should not have "cross chain" features in the base demo.

What I'm proposing is testing only base features here, but for all wallets.
If we remove access to derived wallet in the base demo, the only way to test them would be to re-implement them in the cross-chain demo.

So the proposal is:
- base demo, with optional wallet derivation for testing base features
- cross chain demo, that derives wallets automatically and tests cross-chain features like CCTP

packages/derived-wallet-solana/src/SolanaDerivedWallet.ts
hardsetting: The SIWS standard will make sure the Solana address is displayed.
Then, we should include the Aptos address if applicable (it's not a required field for `StructuredMessage`)

Just for clarity, the SIWS message would look something like this:

Screenshot 2025-03-20 at 12 43 33 PM

where only `message` and `nonce` were specified

packages/derived-wallet-solana/src/SolanaDerivedWallet.ts
hardsetting: > The message needs to be signed by the original wallet private key, i.e solana or evn

Yes we're using the Solana account for signing the SIWS envelope

apps/nextjs-example/src/components/WalletProvider.tsx
0xmaayan: agreed

0xmaayan (approved):
Lets ship it!

PR #514, opened 2025-03-24
[derived-wallet-solana] Return correctly-formed `AbstractAuthData`

hardsetting:
In #504 I just defined a placeholder account authenticator when signing a transaction.
After some [more thinking / digging](https://aptos-org.slack.com/archives/C08DNU2PE9W/p1742609808632769), I came to the conclusion we need to pass the original signing message to the on-chain authentication function in order to perform authentication.

Screenshot 2025-03-24 at 12 38 47 PM

### DIFF
--- a/packages/derived-wallet-solana/src/SolanaDerivedPublicKey.ts
+++ b/packages/derived-wallet-solana/src/SolanaDerivedPublicKey.ts
@@ -5,6 +5,7 @@ import {
   Deserializer,
   Ed25519PublicKey,
   Ed25519Signature,
+  hashValues,
   Serializer,
   VerifySignatureArgs,
 } from '@aptos-labs/ts-sdk';
@@ -12,7 +13,6 @@ import { createSignInMessage as createSolanaSignInMessage } from '@solana/wallet
 import { PublicKey as SolanaPublicKey } from '@solana/web3.js';
 import { createSiwsEnvelopeForAptosStructuredMessage } from './createSiwsEnvelopeForStructuredMessage';
 import { createSiwsEnvelopeForAptosTransaction } from './createSiwsEnvelopeForTransaction';
-
 
 export interface SolanaDerivedPublicKeyParams {
   domain: string;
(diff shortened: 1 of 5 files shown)


packages/derived-wallet-solana/src/SolanaDerivedPublicKey.ts
```diff
@@ -41,7 +40,7 @@ export class SolanaDerivedPublicKey extends AccountPublicKey {
     const serializer = new Serializer();
     serializer.serializeBytes(utf8EncodedDomain);
     serializer.serializeFixedBytes(solanaPublicKeyBytes); // fixed length 32 bytes
-    const accountIdentifier = hashValues([serializer.toUint8Array()]);
+    const accountIdentifier = serializer.toUint8Array();
```
hardsetting, line 43: Not hashing this anymore, so it can be deserialized in the authentication function to obtain the domain and Solana public key

packages/derived-wallet-solana/src/signAptosTransaction.ts
```diff
@@ -50,12 +51,21 @@ export async function signAptosTransactionWithSolana(input: SignAptosTransaction
...
+    serializer.serializeOption(input.rawTransaction.feePayerAddress);
+    const authenticator = serializer.toUint8Array();
+
     return new AccountAuthenticatorAbstraction(
       authenticationFunction,
-      // Not sure what the expected value is here
-      output.signedMessage,
-      signature.bcsToBytes(),
+      transactionHash,
+      authenticator,
```
hardsetting: Passing the hash of the signing message as `digest`, and the concatenation of `signature` and `signingMessage` as `authenticator`, so we have everything we need for on-chain authentication

packages/derived-wallet-solana/src/SolanaDerivedPublicKey.ts
```diff
@@ -54,25 +53,38 @@ export class SolanaDerivedPublicKey extends AccountPublicKey {
...
-        aptosAddress: this._authKey.derivedAddress(),
-        rawTransaction: parsed.rawTransaction,
+        rawTransaction: parsedSigningMessage.rawTransaction,
+        digest: signingMessageDigest,
       });
 
-    const signInMessage = createSignInMessage(siwsInput);
+    // Matching the signature will ensure that the following fields are matching:
+    // - domain
+    // - solanaPublicKey
```
0xmaayan, line 78: this is also the account address
```suggestion
    // - solanaPublicKey (account address)
```

0xmaayan (approved):
LGTM

PR #515, opened 2025-03-24
[derived-wallet-ethereum] Matched structure of Solana package

hardsetting:
Updated Ethereum package to match structure of Solana package:

- exporting `signAptosTransactionWithEthereum` and `signAptosMessageWithEthereum` that are independent from derived wallets
- extracted envelope statement generation into common package
- using `Nonce` field for signing message digest (still to be discussed)

on top of this:
- removed hardcoded ethereum chain id, and added support for any chain id (value is included in signature)

Screenshot 2025-03-24 at 4 31 02 PM

Screenshot 2025-03-24 at 4 32 06 PM


packages/derived-wallet-solana/src/createSiwsEnvelope.ts
```diff
@@ -0,0 +1,54 @@
...
+function createSiwsEnvelope(input: CreateSiwsEnvelopeInput & {
+  statement: string
+}): SolanaSignInInputWithRequiredFields {
+  const { solanaPublicKey, signingMessageDigest, statement } = input;
+  const digestHex = Hex.fromHexInput(signingMessageDigest).toString();
+  return {
+    address: solanaPublicKey.toString(),
+    domain: window.location.host,
+    uri: window.location.origin,
+    nonce: digestHex,
```
0xmaayan, line 24: using a default value in the `nonce` field is a bit weird from a dev POV, since the `signMessage()` the dev uses requires them to provide `message, nonce` (following the `AptosSignMessageInput` standard)

So we ended up requiring the dev to provide a `nonce` but never really use it and use the `digestHex` by default

hardsetting: The user-provided nonce is not displayed, but it's included in the signing message and will affect the digest

The alternative is that we display the user-provided nonce in the `Nonce` field, and the digest somewhere else.

hardsetting: I guess i would be fine with the displayed nonce being different from the provided one,
since the user-provided nonce is typically randomly generated and not diplayed anywhere.
It's mostly to ensure signatures are not replayable, and the check is done automatically in the background.

0xmaayan: > The alternative is that we display the user-provided nonce in the Nonce field, and the digest somewhere else.

We can also not require the end dev to provide a nonce, and have the adapter provider provide the nonce as the transaction hash

0xmaayan (approved):

PR #518, opened 2025-03-25
Keep only minimum required fields for SIWE and SIWS

0xmaayan:
This PR uses the bare minimum required fields for both SIWS and SIWE.
In addition, for SIWE we use the current Aptos chain id instead of the Ethereum chain id in the message

Note:
- While SIWE minimum required fields are `chainId, domain, nonce, uri, version (default to 1)` Metamask adds `issued, network` to the UI message. Network is the `chainId` in `hex` format, so if `chainId : 2` network is `0x2`

Deployed to https://aptos-labs.github.io/aptos-wallet-adapter/nextjs-cross-chain-example/

### DIFF
--- a/packages/derived-wallet-ethereum/src/EIP1193DerivedPublicKey.ts
+++ b/packages/derived-wallet-ethereum/src/EIP1193DerivedPublicKey.ts
@@ -62,13 +62,13 @@ export class EIP1193DerivedPublicKey extends AccountPublicKey {
       return false;
     }
 
-    const { ethereumChainId, issuedAt, siweSignature } = signature;
+    const { chainId, issuedAt, siweSignature } = signature;
     const signingMessageDigest = hashValues([message]);
 
     // Obtain SIWE envelope for the signing message
     const envelopeInput = {
       ethereumAddress: this.ethereumAddress,
-      ethereumChainId,
+      chainId,
       signingMessageDigest,
       issuedAt,
     };
--- a/packages/derived-wallet-ethereum/src/EIP1193DerivedSignature.ts
+++ b/packages/derived-wallet-ethereum/src/EIP1193DerivedSignature.ts
@@ -4,16 +4,16 @@ export class EIP1193DerivedSignature extends Signature {
   static readonly LENGTH = 65;
 
   private readonly _siweSignature: Uint8Array;
-  readonly ethereumChainId: number;
+  readonly chainId: number;
   readonly issuedAt: Date;
 
-  constructor(siweSignature: HexInput, ethereumChainId: number, issuedAt: Date) {
+  constructor(siweSignature: HexInput, chainId: number, issuedAt: Date) {
     super();
     this._siweSignature = Hex.fromHexInput(siweSignature).toUint8Array();
     if (this._siweSignature.length !== EIP1193DerivedSignature.LENGTH) {
       throw new Error('Expected signature length to be 65 bytes');
     }
-    this.ethereumChainId = ethereumChainId;
+    this.chainId = chainId;
     this.issuedAt = issuedAt;
   }
 
@@ -23,15 +23,15 @@ export class EIP1193DerivedSignature extends Signature {
 
   serialize(serializer: Serializer) {
     serializer.serializeFixedBytes(this._siweSignature);
-    serializer.serializeU32AsUleb128(this.ethereumChainId);
+    serializer.serializeU32AsUleb128(this.chainId);
     serializer.serializeU64(this.issuedAt.getTime());
   }
 
   static deserialize(deserializer: Deserializer) {
     const siweSignature = deserializer.deserializeFixedBytes(EIP1193DerivedSignature.LENGTH);
-    const ethereumChainId = deserializer.deserializeUleb128AsU32();
+    const chainId = deserializer.deserializeUleb128AsU32();
     // Number can safely contain a unix timestamp
     const issuedAt = new Date(Number(deserializer.deserializeU64()));
-    return new EIP1193DerivedSignature(siweSignature, ethereumChainId, issuedAt);
+    return new EIP1193DerivedSignature(siweSignature, chainId, issuedAt);
   }
 }
--- a/packages/derived-wallet-ethereum/src/EIP1193DerivedWallet.ts
+++ b/packages/derived-wallet-ethereum/src/EIP1193DerivedWallet.ts
@@ -219,6 +219,7 @@ export class EIP1193DerivedWallet implements AptosWallet {
     return signAptosMessageWithEthereum({
       eip1193Provider: this.eip1193Provider,
       authenticationFunction: this.authenticationFunction,
+      aptosChainId: NetworkToChainId[this.defaultNetwork],
       messageInput: {
         ...input,
         chainId,
@@ -233,6 +234,7 @@ export class EIP1193DerivedWallet implements AptosWallet {
     return signAptosTransactionWithEthereum({
       eip1193Provider: this.eip1193Provider,
       authenticationFunction: this.authenticationFunction,
+      aptosChainId: NetworkToChainId[this.defaultNetwork],
       rawTransaction,
     })
   }
--- a/packages/derived-wallet-ethereum/src/EIP1193DerivedWallet.ts
+++ b/packages/derived-wallet-ethereum/src/EIP1193DerivedWallet.ts
@@ -215,11 +215,10 @@ export class EIP1193DerivedWallet implements AptosWallet {
   // region Signatures
 
   async signMessage(input: AptosSignMessageInput): Promise<UserResponse<AptosSignMessageOutput>> {
-    const chainId = input.chainId ? NetworkToChainId[this.defaultNetwork] : undefined;
+    const chainId = NetworkToChainId[this.defaultNetwork];
     return signAptosMessageWithEthereum({
       eip1193Provider: this.eip1193Provider,
       authenticationFunction: this.authenticationFunction,
-      aptosChainId: NetworkToChainId[this.defaultNetwork],
       messageInput: {
         ...input,
         chainId,
@@ -234,7 +233,6 @@ export class EIP1193DerivedWallet implements AptosWallet {
     return signAptosTransactionWithEthereum({
       eip1193Provider: this.eip1193Provider,
       authenticationFunction: this.authenticationFunction,
-      aptosChainId: NetworkToChainId[this.defaultNetwork],
       rawTransaction,
     })
   }
--- a/packages/derived-wallet-ethereum/src/createSiweEnvelope.ts
+++ b/packages/derived-wallet-ethereum/src/createSiweEnvelope.ts
@@ -9,19 +9,19 @@ import { EthereumAddress } from './shared';
 
 export interface CreateSiweEnvelopeInput {
   ethereumAddress: EthereumAddress;
-  ethereumChainId: number;
+  chainId: number;
   signingMessageDigest: HexInput;
   issuedAt: Date;
 }
 
 function createSiweEnvelope(input: CreateSiweEnvelopeInput & { statement: string }) {
-  const { ethereumAddress, ethereumChainId, signingMessageDigest, issuedAt, statement } = input;
+  const { ethereumAddress, chainId, signingMessageDigest, issuedAt, statement } = input;
   const digestHex = Hex.fromHexInput(signingMessageDigest).toString();
   return createSiweMessage({
     address: ethereumAddress,
     domain: window.location.host,
     uri: window.location.origin,
-    chainId: ethereumChainId,
+    chainId,
     nonce: digestHex,
     statement,
     version: '1',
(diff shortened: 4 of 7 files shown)


packages/derived-wallet-ethereum/src/signAptosTransaction.ts
```diff
@@ -15,12 +15,13 @@ import { EthereumAddress, wrapEthersUserResponse } from './shared';
 export interface SignAptosTransactionWithEthereumInput {
   eip1193Provider: Eip1193Provider | BrowserProvider;
   ethereumAddress?: EthereumAddress;
+  aptosChainId: number;
```
hardsetting: this is already present in the raw transaction, remove

packages/derived-wallet-ethereum/src/signAptosTransaction.ts
```diff
@@ -34,17 +35,15 @@ export async function signAptosTransactionWithEthereum(input: SignAptosTransacti
...
   const signingMessage = generateSigningMessageForTransaction(rawTransaction);
   const signingMessageDigest = hashValues([signingMessage]);
 
   // We need to provide `issuedAt` externally so that we can match it with the signature
   const issuedAt = new Date();
+
   const siweMessage = createSiweEnvelopeForAptosTransaction({
     ethereumAddress,
-    ethereumChainId,
+    chainId: aptosChainId,
```
hardsetting: this needs to be the ethereum chain id, pls revert

packages/derived-wallet-ethereum/src/signAptosTransaction.ts
```diff
@@ -53,7 +52,7 @@ export async function signAptosTransactionWithEthereum(input: SignAptosTransacti
   const response = await wrapEthersUserResponse(ethereumAccount.signMessage(siweMessage));
 
   return mapUserResponse(response, (siweSignature) => {
-    const signature = new EIP1193DerivedSignature(siweSignature, ethereumChainId, issuedAt);
+    const signature = new EIP1193DerivedSignature(siweSignature, aptosChainId, issuedAt);
```
hardsetting: pls revert

packages/derived-wallet-ethereum/src/signAptosMessage.ts
```diff
@@ -65,7 +64,7 @@ export async function signAptosMessageWithEthereum(input: SignAptosMessageWithEt
   const issuedAt = new Date();
   const siweMessage = createSiweEnvelopeForAptosStructuredMessage({
     ethereumAddress,
-    ethereumChainId,
+    chainId: aptosChainId,
```
hardsetting: this needs to be the ethereum chain id, pls revert

packages/derived-wallet-ethereum/src/createSiweEnvelope.ts
```diff
@@ -9,19 +9,19 @@ import { EthereumAddress } from './shared';
...
 
 function createSiweEnvelope(input: CreateSiweEnvelopeInput & { statement: string }) {
-  const { ethereumAddress, ethereumChainId, signingMessageDigest, issuedAt, statement } = input;
+  const { ethereumAddress, chainId, signingMessageDigest, issuedAt, statement } = input;
   const digestHex = Hex.fromHexInput(signingMessageDigest).toString();
   return createSiweMessage({
     address: ethereumAddress,
     domain: window.location.host,
     uri: window.location.origin,
-    chainId: ethereumChainId,
```
hardsetting, line 24: pls revert

packages/derived-wallet-ethereum/src/EIP1193DerivedWallet.ts
```diff
@@ -219,6 +219,7 @@ export class EIP1193DerivedWallet implements AptosWallet {
     return signAptosMessageWithEthereum({
       eip1193Provider: this.eip1193Provider,
       authenticationFunction: this.authenticationFunction,
+      aptosChainId: NetworkToChainId[this.defaultNetwork],
```
hardsetting: pls revert

packages/derived-wallet-ethereum/src/EIP1193DerivedSignature.ts
```diff
@@ -4,16 +4,16 @@ export class EIP1193DerivedSignature extends Signature {
   static readonly LENGTH = 65;
 
   private readonly _siweSignature: Uint8Array;
-  readonly ethereumChainId: number;
```
hardsetting, line 7: pls revert

packages/derived-wallet-ethereum/src/EIP1193DerivedPublicKey.ts
```diff
@@ -59,13 +59,13 @@ export class EIP1193DerivedPublicKey extends AccountPublicKey {
       return false;
     }
 
-    const { ethereumChainId, issuedAt, siweSignature } = signature;
+    const { chainId, issuedAt, siweSignature } = signature;
     const signingMessageDigest = hashValues([message]);
 
     // Obtain SIWE envelope for the signing message
     const envelopeInput = {
       ethereumAddress: this.ethereumAddress,
```
hardsetting, line 70: pls revert

hardsetting (changes requested):
siwe expects an ethereum chain id, why is this being changed?

the aptos chain identifier is available in the statement

0xmaayan:
> siwe expects an ethereum chain id, why is this being changed?
> 
> the aptos chain identifier is available in the statement

The purpose is to sign a message or transaction on the Aptos chain, so it makes sense to show the Aptos chain we intend to submit the transaction to. The chain id should be the chain id to which the session is bound, and in our case we "bound" to the Aptos chain

hardsetting:
>The purpose is to sign a message or transaction on the Aptos chain, so it makes sense to show the Aptos chain we intend to submit the transaction to. The chain id should be the chain id to which the session is bound, and in our case we "bound" to the Aptos chain

The SIWE wrapper is relative to Ethereum and displayed within an Ethereum wallet.
The fields displayed below the statement are relative to Ethereum unless specified.

From ERC-4361:
> `chain-id` REQUIRED. The [EIP-155](https://eips.ethereum.org/EIPS/eip-155) Chain ID to which the session is bound, and the network where Contract Accounts MUST be resolved.

Theoretically, a well-implemented standard should prevent signatures for an ethereum network different from the active one, so if the active network is Ethereum Mainnet (chain id 1), with the approach above signing for Aptos testnet (2) or devnet (147 at the moment) would fail.
In practice, neither Metamask nor Phantom enforce this check at the moment, so we *could* use the `Chain ID` field to store the Aptos Chain Id.

But:
- for structured messages, the chain id is not always available or applicable (`chainId` is optional and you can sign messages that are not tied to a specific network), so what chain id should we pass instead?
- the Aptos blockchain name or id is already displayed in the statement, so there's redundancy

Is the concern that the Chain ID displayed below the statement would somehow confuse the user?

0xmaayan:
> > The purpose is to sign a message or transaction on the Aptos chain, so it makes sense to show the Aptos chain we intend to submit the transaction to. The chain id should be the chain id to which the session is bound, and in our case we "bound" to the Aptos chain
> 
> The SIWE wrapper is relative to Ethereum and displayed within an Ethereum wallet. The fields displayed below the statement are relative to Ethereum unless specified.
> 
> From ERC-4361:
> 
> > `chain-id` REQUIRED. The [EIP-155](https://eips.ethereum.org/EIPS/eip-155) Chain ID to which the session is bound, and the network where Contract Accounts MUST be resolved.
> 
> Theoretically, a well-implemented standard should prevent signatures for an ethereum network different from the active one, so if the active network is Ethereum Mainnet (chain id 1), with the approach above signing for Aptos testnet (2) or devnet (147 at the moment) would fail. In practice, neither Metamask nor Phantom enforce this check at the moment, so we _could_ use the `Chain ID` field to store the Aptos Chain Id.
> 
> But:
> 
> * for structured messages, the chain id is not always available or applicable (`chainId` is optional and you can sign messages that are not tied to a specific network), so what chain id should we pass instead?
> * the Aptos blockchain name or id is already displayed in the statement, so there's redundancy
> 
> Is the concern that the Chain ID displayed below the statement would somehow confuse the user?

Yes, I understand that EIP 155 is relative to Ethereum. But we are trying to "trick" the system by using evm/solana wallets to interact with the Aptos network, so in our context the `Chain ID to which the session is bound, and the network where Contract Accounts MUST be resolved` is actually the Aptos network because we are providing users the option to use x-chain wallets to interact with the Aptos network.

I think it can be confusing for the user to see an evm chainId in the message but it is actually been submitted to an aptos chainId.

> * for structured messages, the chain id is not always available or applicable (`chainId` is optional and you can sign messages that are not tied to a specific network), so what chain id should we pass instead?

We can always provide it in the adapter. And if EVM standard is not to provide a chainId to a structured message, then we dont provide it.

I am not saying we must provide chainId, but I am saying if we already provide it, it should be an Aptos chainId.

> * the Aptos blockchain name or id is already displayed in the statement, so there's redundancy

I'd argue this is actually confusing. The Aptos blockchain name or id is displayed in the statement, but then also an ethereum chain id is displayed.

hardsetting:
My preference would be to follow the standard to avoid problems in the future (e.g. wallet complaining about mismatched chain id), but I don't feel strongly enough to push it so as long as everything works let's go with it.
Most times the Chain ID will be matching anyways :P (bot mainnets are chain id 1)

Let's just maybe remove redundancy in parameters

packages/derived-wallet-ethereum/src/EIP1193DerivedWallet.ts
```diff
@@ -233,6 +234,7 @@ export class EIP1193DerivedWallet implements AptosWallet {
     return signAptosTransactionWithEthereum({
       eip1193Provider: this.eip1193Provider,
       authenticationFunction: this.authenticationFunction,
+      aptosChainId: NetworkToChainId[this.defaultNetwork],
```
hardsetting: The chain ID is already available as field of `RawTransaction`, pls remove

0xmaayan:
> My preference would be to follow the standard to avoid problems in the future (e.g. wallet complaining about mismatched chain id), but I don't feel strongly enough to push it so as long as everything works let's go with it. Most times the Chain ID will be matching anyways :P (bot mainnets are chain id 1)
> 

SGTM.

> Let's just maybe remove redundancy in parameters

wdym?

packages/derived-wallet-ethereum/src/EIP1193DerivedWallet.ts
```diff
@@ -219,6 +219,7 @@ export class EIP1193DerivedWallet implements AptosWallet {
     return signAptosMessageWithEthereum({
       eip1193Provider: this.eip1193Provider,
       authenticationFunction: this.authenticationFunction,
+      aptosChainId: NetworkToChainId[this.defaultNetwork],
       messageInput: {
         ...input,
         chainId,
```
hardsetting: Let's do something like this maybe?

```typescript
const chainId = NetworkToChainId[this.defaultNetwork];
return signAptosMessageWithEthereum({
  eip1193Provider: this.eip1193Provider,
  authenticationFunction: this.authenticationFunction,
  chainId,
  messageInput: {
    ...input,
    chainId: input.chainId ? chainId : undefined,
  },
});
```

packages/derived-wallet-ethereum/src/signAptosMessage.ts
```diff
@@ -19,12 +19,13 @@ export interface StructuredMessageInputWithChainId extends StructuredMessageInpu
 export interface SignAptosMessageWithEthereumInput {
   eip1193Provider: Eip1193Provider | BrowserProvider;
   ethereumAddress?: EthereumAddress;
+  aptosChainId: number;
```
hardsetting: rename to `chainId` for consistency?

hardsetting:
>wdym?

just added a few comments

packages/derived-wallet-ethereum/src/EIP1193DerivedWallet.ts
hardsetting: Alternatively we could just force the message to include the chain id, that might be simpler.

```typescript
const chainId = NetworkToChainId[this.defaultNetwork];
return signAptosMessageWithEthereum({
  eip1193Provider: this.eip1193Provider,
  authenticationFunction: this.authenticationFunction,
  messageInput: {
    ...input,
    chainId,
  },
});
```

and then change the chainId to required in the params for `signAptosMessageWithEthereum`

hardsetting: whichever you prefer

0xmaayan:
> > wdym?
> 
> just added a few comments

Updated!

Just realized we would need to manually handle devnet, so it fetches the current devnet chain id - can address it in a next PR

packages/derived-wallet-ethereum/src/EIP1193DerivedSignature.ts
```diff
@@ -4,16 +4,16 @@ export class EIP1193DerivedSignature extends Signature {
...
+  readonly chainId: number;
   readonly issuedAt: Date;
 
-  constructor(siweSignature: HexInput, ethereumChainId: number, issuedAt: Date) {
+  constructor(siweSignature: HexInput, chainId: number, issuedAt: Date) {
```
BriungRi, line 10: which `chainId`? Could be nice to be called `aptosChainId` if it is the aptos one

BriungRi (approved):

hardsetting (approved):
ty!!

PR #520, opened 2025-03-25
Use the Aptos Labs bot secret with changeset

0xmaayan:
Start using the Aptos Labs bot secret for changeset flow
More context https://aptos-org.slack.com/archives/C03EP0XM99D/p1740093496332859


hardsetting (changes requested):
You're swapping an access token for an app private key

you can refer to this workflow for reference on how to obtain a token

https://github.com/aptos-labs/aptos-connect/blob/main/.github/workflows/publish-packages.yml

hardsetting (approved):

PR #522, opened 2025-03-26
Bump sdk to 1.37.1

heliuchuan:


kent-white (approved):

packages/derived-wallet-ethereum/src/EIP1193DerivedPublicKey.ts
```diff
@@ -84,6 +87,10 @@ export class EIP1193DerivedPublicKey extends AccountPublicKey {
     return recoveredAddress === this.ethereumAddress;
   }
 
+  async verifySignatureAsync(args: { aptosConfig: AptosConfig, message: HexInput, signature: Signature }): Promise<boolean> {
```
0xmaayan, line 90: I am surprised it lets you pass  `aptosConfig: AptosConfig` when `this.verifySignature` does not expect this property

0xmaayan (commented):
Please run changeset

packages/derived-wallet-ethereum/src/EIP1193DerivedPublicKey.ts
heliuchuan: Yeah, if you explicitly define it it'll complain but if you pass the whole object its chill

0xmaayan: can we be more specific here and only pass the expected props?
```
const { message, signature } = args
this.verifySignature({message, signature})
```

PR #523, opened 2025-03-26
Introduce x chain wallets logic in the adapter provider

0xmaayan:
Introducing x-chain wallets related logic in the react adapter provider for a better dev exp

### DIFF
--- a/apps/nextjs-x-chain/src/app/page.tsx
+++ b/apps/nextjs-x-chain/src/app/page.tsx
@@ -18,19 +18,17 @@ import {
   AdapterWallet,
   AptosChangeNetworkOutput,
   NetworkInfo,
+  OriginWalletDetails,
   isAptosNetwork,
   useWallet,
 } from "@aptos-labs/wallet-adapter-react";
 import { init as initTelegram } from "@telegram-apps/sdk";
 import { AlertCircle } from "lucide-react";
 import Image from "next/image";
-import { PublicKey as SolanaPublicKey } from "@solana/web3.js";
 // Imports for registering a browser extension wallet plugin on page load
 import { MyWallet } from "@/utils/standardWallet";
 import { registerWallet } from "@aptos-labs/wallet-standard";
-import { SolanaDerivedWallet } from "@aptos-labs/derived-wallet-solana";
 import { CCTPTransfer } from "@/components/CCTPTransfer";
-import { EIP1193DerivedWallet } from "@aptos-labs/derived-wallet-ethereum";
 
 // Example of how to register a browser extension wallet plugin.
 // Browser extension wallets should call registerWallet once on page load.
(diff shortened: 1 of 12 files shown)


packages/wallet-adapter-react/src/WalletProvider.tsx
```diff
@@ -392,6 +407,57 @@ export const AptosWalletAdapterProvider: FC<AptosWalletProviderProps> = ({
...
+  ): Promise<EVMWalletDetails>;
+  function getOriginWalletDetails(
+    wallet: AdapterWallet
+  ): Promise<AptosWalletDetails>;
+
+  // Implementation
+  async function getOriginWalletDetails(
+    wallet: AdapterWallet
+  ): Promise<OriginWalletDetails> {
+    if (isSolanaDerivedWallet(wallet)) {
```
hardsetting, line 428: why not using `instanceof` directly here?

packages/wallet-adapter-react/src/WalletProvider.tsx
```diff
@@ -392,6 +407,57 @@ export const AptosWalletAdapterProvider: FC<AptosWalletProviderProps> = ({
...
+  // Function overloads
+  function getOriginWalletDetails(
+    wallet: SolanaDerivedWallet
+  ): Promise<SolanaWalletDetails>;
+  function getOriginWalletDetails(
+    wallet: EIP1193DerivedWallet
+  ): Promise<EVMWalletDetails>;
+  function getOriginWalletDetails(
+    wallet: AdapterWallet
+  ): Promise<AptosWalletDetails>;
```
hardsetting: is this needed?

apps/nextjs-x-chain/src/app/page.tsx
```diff
@@ -272,34 +253,40 @@ function WalletConnection({
...
+                          originWalletDetails?.publicKey?.toString() ??
+                          "Not Present"
+                        }
+                        isCorrect={!!originWalletDetails?.publicKey}
+                      />
+                    ),
+                  },
+                ]}
+              />
+            </div>
```
hardsetting: how would you feel about something like:

```typescript
if (isSolanaDerivedWallet(wallet)) {
  // show solana-specific UI
} else if (isEIP1193DerivedWallet(wallet) {
  // show etherem-specific UI
}
```

the reason being, "public keys" are not applicable to ethereum, and the public key "is" the address for Solana
not sure if it's "useful" to try and combine them into the same data type

hardsetting: Or alternatively, just show the address :)

apps/nextjs-x-chain/src/components/CCTPTransfer.tsx
```diff
@@ -109,7 +121,7 @@ export function CCTPTransfer({
     if (!originWalletDetails) return;
     const fetchWalletUsdcBalance = async () => {
       const balance = await crossChainCore.getWalletUSDCBalance(
```
hardsetting, line 123: USDC balance on source chain?

hardsetting (approved):
Looks good! Just a few comments

packages/wallet-adapter-react/src/WalletProvider.tsx
hardsetting: Yeah actually I would just return undefined, if there's no "source" wallet

apps/nextjs-x-chain/src/components/CCTPTransfer.tsx
0xmaayan: are you suggesting a rename? or asking what this function does?

packages/wallet-adapter-react/src/WalletProvider.tsx
0xmaayan: from the dev POV, I'd want them to call this function with a wallet
```
const details = getOriginWalletDetails(wallet)
```
and not handle the wallet type in their code, for example not need to do something like
```
if(isSolana(wallet) || isEthereum(wallet)){
const details = getOriginWalletDetails(wallet)
}
```

packages/wallet-adapter-react/src/WalletProvider.tsx
0xmaayan: we can, but we already have the helper function that does exactly that

apps/nextjs-x-chain/src/app/page.tsx
0xmaayan: good call. maybe just display it in the "wallet info" section

PR #528, opened 2025-03-27
Fetch the devnet chain id when x-chain wallet is configured to use th…

0xmaayan:
…e devnet

### DIFF
--- a/packages/derived-wallet-base/src/utils.ts
+++ b/packages/derived-wallet-base/src/utils.ts
@@ -1,4 +1,4 @@
-import { AccountPublicKey } from '@aptos-labs/ts-sdk';
+import { AccountPublicKey, Aptos } from '@aptos-labs/ts-sdk';
 import { AccountInfo } from '@aptos-labs/wallet-standard';
 
 export function accountInfoFromPublicKey(publicKey: AccountPublicKey) {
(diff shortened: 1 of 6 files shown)


packages/derived-wallet-base/src/utils.ts
```diff
@@ -11,3 +11,11 @@ export function accountInfoFromPublicKey(publicKey: AccountPublicKey) {
 export function isNullCallback(callback: Function) {
   return '_isNull' in callback && callback._isNull === true;
 }
+
+/**
+ * Helper function to fetch Devnet chain id
+ */
+export const fetchDevnetChainId = async (): Promise<number> => {
+  const aptos = new Aptos(); // default to devnet
```
hardsetting, line 19: oh that's surprising, would think it defaults to mainnet :D

packages/derived-wallet-ethereum/src/EIP1193DerivedWallet.ts
```diff
@@ -215,7 +215,7 @@ export class EIP1193DerivedWallet implements AptosWallet {
   // region Signatures
 
   async signMessage(input: AptosSignMessageInput): Promise<UserResponse<AptosSignMessageOutput>> {
-    const chainId = NetworkToChainId[this.defaultNetwork];
+    const chainId = this.defaultNetwork === Network.DEVNET ? await fetchDevnetChainId() :NetworkToChainId[this.defaultNetwork];
```
hardsetting, line 218: :(

hardsetting (approved):

PR #535, opened 2025-03-31
DAA Improvements to work with e2e flow

0xmaayan:
- Add the accountIdentity when creating the AccountAuthenticatorAbstraction in the derived wallet
- Fix DAA address computation
- Use same domain in message and in the wallet class variable
- Fix signature sent to chain
- Bump ts-sdk version

### DIFF
--- a/packages/derived-wallet-base/src/abstraction.ts
+++ b/packages/derived-wallet-base/src/abstraction.ts
@@ -1,15 +1,84 @@
-import { AuthenticationKey, DerivableAbstractedAccount } from '@aptos-labs/ts-sdk';
+import {
+  AccountAddress,
+  AuthenticationKey,
+  Deserializer,
+  hashValues,
+  isValidFunctionInfo,
+  Serializable,
+  Serializer,
+} from "@aptos-labs/ts-sdk";
+
+/**
+ * The domain separator used to calculate the DAA account address.
+ */
+export const ADDRESS_DOMAIN_SEPARATOR = 5;
 
 /**
  * @param functionInfo - The authentication function
- * @param accountIdentifier - The account identity
- * @returns The account address
+ * @param identity - The account identity. Typically it is the wallet public key or account address
+ * @param domain - The dapp domain
+ * @returns The account address authentication key
  */
-export function computeDomainAuthenticationKey(functionInfo: string, accountIdentifier: Uint8Array): AuthenticationKey {
-  try{
-    const add = DerivableAbstractedAccount.computeAccountAddress(functionInfo, accountIdentifier);
-    return new AuthenticationKey({ data:add });
+export function computeDerivableAuthenticationKey(
+  functionInfo: string,
+  identity: string,
+  domain: string
+): AuthenticationKey {
+  try {
+    // TODO: move it to the ts-sdk as the general computation function once we are finalized on the on-chain auth function
+    if (!isValidFunctionInfo(functionInfo)) {
+      throw new Error(`Invalid authentication function ${functionInfo}`);
+    }
+    const [moduleAddress, moduleName, functionName] = functionInfo.split("::");
+
+    // Serialize and append the function info
+    const serializer = new Serializer();
+    AccountAddress.fromString(moduleAddress).serialize(serializer);
+    serializer.serializeStr(moduleName);
+    serializer.serializeStr(functionName);
+
+    // Serialize and append the abstract public key
+    const s2 = new Serializer();
+    const abstractPublicKey = new DerivableAbstractPublicKey(identity, domain);
+    s2.serializeBytes(abstractPublicKey.bcsToBytes());
+
+    // Append the domain separator
+    const domainSeparator = new Uint8Array([ADDRESS_DOMAIN_SEPARATOR]);
+
+    const data = hashValues([
+      serializer.toUint8Array(),
+      s2.toUint8Array(),
+      domainSeparator,
+    ]);
+
+    return new AuthenticationKey({ data });
   } catch (e) {
-    throw new Error(`Error computing domain authentication key: ${e}`);
+    throw `Error computing domain authentication key: ${e}`;
+  }
+}
+
+/**
+ * The derivable abstract public key of the DAA account.
+ * 
+ * @param identity - The identity of the account. Typically it is the wallet public key or account address
+ * @param domain - The dapp domain
+ */
+export class DerivableAbstractPublicKey extends Serializable {
+  constructor(
+    public identity: string,
+    public domain: string
+  ) {
+    super();
+  }
+
+  serialize(serializer: Serializer): void {
+    serializer.serializeStr(this.identity);
+    serializer.serializeStr(this.domain);
+  }
+
+  static deserialize(deserializer: Deserializer): DerivableAbstractPublicKey {
+    const identity = deserializer.deserializeStr();
+    const domain = deserializer.deserializeStr();
+    return new DerivableAbstractPublicKey(identity, domain);
   }
 }
--- a/packages/derived-wallet-base/src/abstraction.ts
+++ b/packages/derived-wallet-base/src/abstraction.ts
@@ -1,43 +1,15 @@
-import { AccountAddress, AuthenticationKey, hashValues, isValidFunctionInfo, Serializer } from '@aptos-labs/ts-sdk';
+import { AuthenticationKey, DerivableAbstractedAccount } from '@aptos-labs/ts-sdk';
 
 /**
- * The domain separator used to calculate the DAA account address.
- */
-export const ADDRESS_DOMAIN_SEPARATOR = 5;
-
-/**
- * Compute the account address of the DAA
- * The DAA account address is computed by hashing the function info and the account identity
- * and appending the domain separator (5)
- *
  * @param functionInfo - The authentication function
  * @param accountIdentifier - The account identity
  * @returns The account address
  */
-export function computeDomainAuthenticationKey(functionInfo: string, accountIdentifier: Uint8Array) {
-  if (!isValidFunctionInfo(functionInfo)) {
-    throw new Error(`Invalid authentication function ${functionInfo}`);
+export function computeDomainAuthenticationKey(functionInfo: string, accountIdentifier: Uint8Array): AuthenticationKey {
+  try{
+    const add = DerivableAbstractedAccount.computeAccountAddress(functionInfo, accountIdentifier);
+    return new AuthenticationKey({ data:add });
+  } catch (e) {
+    throw new Error(`Error computing domain authentication key: ${e}`);
   }
-  const [moduleAddress, moduleName, functionName] = functionInfo.split("::");
-
-  // Serialize and append the function info
-  const serializer = new Serializer();
-  AccountAddress.fromString(moduleAddress).serialize(serializer);
-  serializer.serializeStr(moduleName);
-  serializer.serializeStr(functionName);
-
-  // Serialize and append the account identity
-  const s2 = new Serializer();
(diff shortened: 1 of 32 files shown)


packages/derived-wallet-ethereum/src/signAptosTransaction.ts
```diff
@@ -55,6 +56,9 @@ export async function signAptosTransactionWithEthereum(input: SignAptosTransacti
   return mapUserResponse(response, (siweSignature) => {
     const signature = new EIP1193DerivedSignature(siweSignature, chainId, issuedAt);
     const authenticator = signature.bcsToBytes();
+
+    const accountIdentity = `${ethereumAddress.toString()}:${domain}`
```
hardsetting: where is this used?

packages/derived-wallet-solana/src/extendSolanaWallet.ts
```diff
@@ -40,6 +40,7 @@ export function extendSolanaWallet(
     solanaWallet,
     authenticationFunction,
     rawTransaction,
+    domain:""
```
hardsetting: removed hardcoded value

packages/derived-wallet-solana/src/signAptosTransaction.ts
```diff
@@ -58,10 +59,13 @@ export async function signAptosTransactionWithSolana(input: SignAptosTransaction
     serializer.serializeOption(input.rawTransaction.feePayerAddress);
     const authenticator = serializer.toUint8Array();
 
+    const accountIdentity = `${solanaPublicKey.toBase58()}:${domain}`
+    
     return new AccountAuthenticatorAbstraction(
       authenticationFunction,
       signingMessageDigest,
       authenticator,
+      new TextEncoder().encode(accountIdentity)
```
hardsetting: i think this is already being computed as field of the derived public key

could we try and reuse that if possible?

packages/derived-wallet-ethereum/src/signAptosTransaction.ts
0xmaayan: removed

packages/derived-wallet-solana/src/extendSolanaWallet.ts
0xmaayan: done

packages/derived-wallet-base/src/abstraction.ts
```diff
@@ -1,43 +1,15 @@
...
  */
-export function computeDomainAuthenticationKey(functionInfo: string, accountIdentifier: Uint8Array) {
-  if (!isValidFunctionInfo(functionInfo)) {
-    throw new Error(`Invalid authentication function ${functionInfo}`);
+export function computeDomainAuthenticationKey(functionInfo: string, accountIdentifier: Uint8Array): AuthenticationKey {
+  try{
+    const add = DerivableAbstractedAccount.computeAccountAddress(functionInfo, accountIdentifier);
+    return new AuthenticationKey({ data:add });
+  } catch (e) {
+    throw new Error(`Error computing domain authentication key: ${e}`);
```
hardsetting: Why are we catching and throwing a different error?
We lose information about the stack trace etc..

How about letting the consumer catch the underlying error instead?

packages/derived-wallet-base/src/abstraction.ts
```diff
@@ -1,43 +1,15 @@
...
  * @param accountIdentifier - The account identity
  * @returns The account address
  */
-export function computeDomainAuthenticationKey(functionInfo: string, accountIdentifier: Uint8Array) {
-  if (!isValidFunctionInfo(functionInfo)) {
-    throw new Error(`Invalid authentication function ${functionInfo}`);
+export function computeDomainAuthenticationKey(functionInfo: string, accountIdentifier: Uint8Array): AuthenticationKey {
+  try{
+    const add = DerivableAbstractedAccount.computeAccountAddress(functionInfo, accountIdentifier);
+    return new AuthenticationKey({ data:add });
```
hardsetting: Unrelated to this PR: I would suggest possibly renaming the original function from the SDK.
I think "authentication key" would be more apt.

For context, the address of an account always matches the authentcation key of its original public key.
It only differs when the authentication key is rotated.

packages/derived-wallet-base/src/envelope.ts
```diff
@@ -16,8 +16,9 @@ function getChainName(chainId: number) {
       return network;
     }
   }
-  // Otherwise return a chain id descriptor
-  return `chainId: ${chainId}`;
+  // Otherwise return null. Should really only happen for devnet, and on chain
+  // the auth function does not expect a chain id for devnet
+  return null;
```
hardsetting: Can we revert this and make the on-chain function match?

I don't see a reason against making this work for all chains, other than the fact that is currently not implemented in the on-chain authentication function at the moment.
I had left a comment in that PR not sure if I got a response on it, lemme check

hardsetting: cc @BriungRi

packages/derived-wallet-base/src/envelope.ts
```diff
@@ -56,8 +56,22 @@ export function createTransactionStatement(rawTransaction: AnyRawTransaction) {
...
+ * Note: If the chainId is devnet, return an empty string as the on-chain auth function 
+ * does not expect a chain id for devnet.
+ */
+export function getAptosChainSuffix(chainId: number | undefined) {
+  if (!chainId) {
+    return '';
+  }
+  const chainName = getChainName(chainId);
+  return chainName ? ` (${chainName})` : '';
+}
```
hardsetting: can we consider reverting?

packages/derived-wallet-solana/src/SolanaDerivedWallet.ts
```diff
@@ -203,14 +203,14 @@ export class SolanaDerivedWallet implements AptosWallet {
...
 
   async signMessage(input: AptosSignMessageInput): Promise<UserResponse<AptosSignMessageOutput>> {
-    const chainId = input.chainId ? this.defaultNetwork === Network.DEVNET ? await fetchDevnetChainId() : NetworkToChainId[this.defaultNetwork] : undefined;
     return signAptosMessageWithSolana({
       solanaWallet: this.solanaWallet,
       authenticationFunction: this.authenticationFunction,
       messageInput: {
         ...input,
-        chainId,
+        chainId:  NetworkToChainId[this.defaultNetwork],
```
hardsetting:
```suggestion
        chainId: NetworkToChainId[this.defaultNetwork],
```

we should probably set up prettier / eslint at some point

packages/derived-wallet-solana/src/shared.ts
```diff
@@ -2,7 +2,7 @@ import { makeUserApproval, makeUserRejection } from '@aptos-labs/derived-wallet-
 import { UserResponse } from '@aptos-labs/wallet-standard';
 import { WalletError } from '@solana/wallet-adapter-base';
 
-export const defaultAuthenticationFunction = '0x7::solana::authenticate';
+export const defaultAuthenticationFunction = '0x7::daa_siws_phantom::authenticate';
```
hardsetting: Afaik this is not specific to Phantom right? Why is that in the name

packages/derived-wallet-solana/src/signAptosTransaction.ts
```diff
@@ -36,32 +36,26 @@ export async function signAptosTransactionWithSolana(input: SignAptosTransaction
...
-
     const signature = new Ed25519Signature(output.signature);
 
-    const serializer = new Serializer();
-    serializer.serialize(signature);
-    serializer.serialize(input.rawTransaction.rawTransaction);
-    serializer.serializeVector(input.rawTransaction.secondarySignerAddresses ?? []);
-    serializer.serializeOption(input.rawTransaction.feePayerAddress);
-    const authenticator = serializer.toUint8Array();
+    const accountIdentity = `${solanaPublicKey.toBase58()}${domain}`
```
hardsetting: As mentioned in a previous comment, this logic is already present in `SolanaDerivedPublicKey`.
How about extract it into a function and use that in both places?
```
function getSolanaDerivedAccountIdentity(solanaPublicKey, domain): Uint8Array {
  return new TextEncoder(`${solanaPublicKey.toBase58()}${domain}`);
}
```

packages/derived-wallet-solana/src/SolanaDerivedPublicKey.ts
```diff
@@ -39,17 +39,10 @@ export class SolanaDerivedPublicKey extends AccountPublicKey {
...
 
-    const utf8EncodedDomain = new TextEncoder().encode(domain);
-    const solanaPublicKeyBytes = solanaPublicKey.toBytes();
-
-    const serializer = new Serializer();
-    serializer.serializeBytes(utf8EncodedDomain);
-    serializer.serializeFixedBytes(solanaPublicKeyBytes); // fixed length 32 bytes
-    const accountIdentifier = serializer.toUint8Array();
-
+    const accountIdentifier = `${solanaPublicKey.toBase58()}${domain}`;
```
hardsetting: I'm ok with dropping BCS if it simplifies the on-chain de-serialization, but as-is this [is a bit inefficient](https://chatgpt.com/share/67eed6d6-dab4-800f-b2fa-a4d906c15892) since we're UTF-8 encoding a B58 encoded bytes array (double encoding).

Can we maybe concatenate the public key bytes with the UTF-8 encoded domain?
Or revert to BCS since we do have an on-chain BCS module for deserialization.

I'm guessing this was done for simplicity, so we can just utf8-decode and slice the identifier into its components,
but the authenticator info will (I think) live on chain with the submitted transaction, so reducing its size would be good.

cc @BriungRi

packages/derived-wallet-base/src/envelope.ts
hardsetting: Ok [caught up](https://github.com/aptos-labs/aptos-core/pull/16236#discussion_r2027583199) in the other thread, whether we display something or nothing at all is up to y'all.

Just one nit, can we return `undefined` instead of `null`?

packages/derived-wallet-base/src/envelope.ts
hardsetting: Could we move this between `getChainName` and the other functions?

packages/derived-wallet-solana/src/SolanaDerivedPublicKey.ts
hardsetting: Oh i see a [similar conversation](https://github.com/aptos-labs/aptos-core/pull/16236#discussion_r2018776148) is happening in the other PR lol

packages/derived-wallet-solana/src/shared.ts
0xmaayan: It is specific to phantom. From our testing wallets can generate the message in different formats, for example Nightly and Metamask on EVM. For Solana, Phantom is the only wallet that actually supports `signInWith` and when other wallets implement it we can't be sure how they will generate the message

see https://github.com/aptos-labs/aptos-core/pull/16236#discussion_r2019455011

packages/derived-wallet-base/src/abstraction.ts
0xmaayan: tbh, authentication key is a bit confusing. I understand the technical side and why it makes sense but I am not completely sure devs really understand what authentication key is for and its meaning.
IMO, account address is a more understandable and used term. I would actually say we should rename it in the wallet adapter.

Also, we might in the future encourage wallets to rotate their auth key to be able to use it in a native Aptos wallet.

hardsetting (approved):

PR #538, opened 2025-04-05
feat: Added support for MSafe Wallet

jovicheng:
feat: Added support for MSafe Wallet

### DIFF
--- a/.changeset/late-cities-dress.md
+++ b/.changeset/late-cities-dress.md
@@ -1,5 +1,5 @@
 ---
-"@aptos-labs/wallet-adapter-core": major
+"@aptos-labs/wallet-adapter-core": minor
 ---
 
 Added support for MSafe Wallet
--- a/.changeset/late-cities-dress.md
+++ b/.changeset/late-cities-dress.md
@@ -1,0 +1,5 @@
+---
+"@aptos-labs/wallet-adapter-core": major
+---
+
+Added support for MSafe Wallet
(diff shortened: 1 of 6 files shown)


.changeset/late-cities-dress.md
```diff
@@ -0,0 +1,5 @@
+---
+"@aptos-labs/wallet-adapter-core": major
```
0xmaayan: This should not be a major update, please change to minor

packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -1050,13 +1054,15 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
       }
 
       const aptosConfig = getAptosConfig(this._network, this._dappConfig);
-      const signingMessage = new TextEncoder().encode(response.args.fullMessage);
+      const signingMessage = new TextEncoder().encode(
```
0xmaayan: Please revert formatting

packages/wallet-adapter-core/src/sdkWallets.ts
```diff
@@ -50,6 +51,15 @@ export function getSDKWallets(dappConfig?: DappConfig) {
     sdkWallets.push(new DevTWallet() as any);
   }
 
+  if (dappConfig?.network) {
```
0xmaayan, line 54: Whats the reason for this check? It doesn't look like it is a required prop for mSafe

jovicheng: image
`network` is required prop for MSafeConfig

0xmaayan (approved):

PR #553, opened 2025-04-16
[example] Fix nextjs example Sponsor txn error 

WGB5445:
Original: Existing process uses multiple wallets to separately sign txn and feepayer transactions. However, due to adaptor and petra coordination issues, having both signatures simultaneously is difficult, leading to verification failures.

Current: Now using a local random account for on-chain operations, with the wallet paying gas and signing the feepayer transaction to complete the flow.

### DIFF
--- a/apps/nextjs-example/src/components/transactionFlows/Sponsor.tsx
+++ b/apps/nextjs-example/src/components/transactionFlows/Sponsor.tsx
@@ -1,6 +1,7 @@
 import { aptosClient, isSendableNetwork } from "@/utils";
 import {
   Account,
+  AccountAddress,
   AccountAuthenticator,
   AnyRawTransaction,
 } from "@aptos-labs/ts-sdk";
@@ -10,8 +11,6 @@ import { TransactionHash } from "../TransactionHash";
 import { Button } from "../ui/button";
 import { Card, CardContent, CardHeader, CardTitle } from "../ui/card";
 import { useToast } from "../ui/use-toast";
-
-const APTOS_COIN = "0x1::aptos_coin::AptosCoin";
 
 export function Sponsor() {
   const { toast } = useToast();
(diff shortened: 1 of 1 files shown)


apps/nextjs-example/src/components/transactionFlows/Sponsor.tsx
```diff
@@ -25,40 +24,46 @@ export function Sponsor() {
...
-      sender: account.address,
+      sender: sender.accountAddress,
       withFeePayer: true,
       data: {
-        function: "0x1::coin::transfer",
-        typeArguments: [APTOS_COIN],
-        functionArguments: [account.address.toString(), 1], // 1 is in Octas
+        function: "0x1::resource_account::create_resource_account",
+        typeArguments: [],
+        functionArguments: ["My Resource Account", AccountAddress.from("0x0").toUint8Array()],
```
0xmaayan: why not keep it `account.address.toString()`?  I find it more readable freindly

apps/nextjs-example/src/components/transactionFlows/Sponsor.tsx
```diff
@@ -25,40 +24,46 @@ export function Sponsor() {
...
       network
     ).transaction.build.simple({
-      sender: account.address,
+      sender: sender.accountAddress,
       withFeePayer: true,
       data: {
-        function: "0x1::coin::transfer",
-        typeArguments: [APTOS_COIN],
-        functionArguments: [account.address.toString(), 1], // 1 is in Octas
+        function: "0x1::resource_account::create_resource_account",
```
0xmaayan, line 43: why are we changing the transaction to submit?

Can we keep it as a transfer transaction, where the sender is the connected wallet, the recipient is the connected wallet, but the fee payer is the local generated account?

apps/nextjs-example/src/components/transactionFlows/Sponsor.tsx
```diff
@@ -101,6 +100,7 @@ export function Sponsor() {
         senderAuthenticator: senderAuthenticator,
         feePayerAuthenticator: feepayerAuthenticator,
       });
+      console.log(response);
```
0xmaayan: remove

apps/nextjs-example/src/components/transactionFlows/Sponsor.tsx
WGB5445: I understand your point. The biggest issue with this approach is that we need to step outside the flow to airdrop test tokens to the account. If this is done too many times, the faucet API might return a 429 error, preventing us from receiving more test tokens. Additionally, the processes for testnet and devnet are not consistent.

What we actually hope to do here is to create a temporary account and have the developer’s own wallet address cover the gas fees. This way, the flow is more likely to work repeatedly without issues.

0xmaayan: so with this change:
- The connected wallet is the fee payer
- The main signer is just a random local generated account?

0xmaayan (approved):

apps/nextjs-example/src/components/transactionFlows/Sponsor.tsx
WGB5445: Yes

PR #560, opened 2025-04-23
Add Solana sign message fallback support + modify message format

0xmaayan:
Add support to fallback to the wallet `signMessage` function, if SIWS is not available. The message format is identical to the one is being used in SIWS so we maintain and support only one message format.

If the wallet supports SIWS, a domain verification will apply. In the case the wallet does not support SIWS the adapter uses `signMessage` function and therefore no domain verification will apply.

This is to balance the need to support many wallets, maintain a one message format in the adapter and on chain and secure users with domain verification when available.

Additionally, this PR modifies the message format to include a warning message, following https://github.com/aptos-labs/aptos-core/pull/16439

### DIFF
--- a/packages/derived-wallet-base/src/envelope.ts
+++ b/packages/derived-wallet-base/src/envelope.ts
@@ -9,7 +9,7 @@ import { StructuredMessage } from './StructuredMessage';
 /**
  * Attempt to convert the specified chainId into a human-readable identifier.
  */
-export function getChainName(chainId: number) {
+function getChainName(chainId: number) {
   // Obtain the network name if available
   for (const [network, otherChainId] of Object.entries(NetworkToChainId)) {
     if (otherChainId === chainId) {
--- a/packages/derived-wallet-base/src/envelope.ts
+++ b/packages/derived-wallet-base/src/envelope.ts
@@ -59,5 +59,5 @@ export function createTransactionStatement(rawTransaction: AnyRawTransaction) {
   const chainName = getChainName(chainId);
   const onAptosChain = ` on Aptos blockchain (${chainName})`;
 
-  return `To execute transaction${humanReadableEntryFunction}${onAptosChain}.`;
+  return `Please confirm you explicitly initiated this request from ${window.location.host}. You are approving to execute transaction${humanReadableEntryFunction}${onAptosChain}.`;
 }
--- a/packages/derived-wallet-base/src/envelope.ts
+++ b/packages/derived-wallet-base/src/envelope.ts
@@ -9,7 +9,7 @@ import { StructuredMessage } from './StructuredMessage';
 /**
  * Attempt to convert the specified chainId into a human-readable identifier.
  */
-function getChainName(chainId: number) {
+export function getChainName(chainId: number) {
   // Obtain the network name if available
   for (const [network, otherChainId] of Object.entries(NetworkToChainId)) {
     if (otherChainId === chainId) {
--- a/packages/derived-wallet-solana/src/createSiwsEnvelope.ts
+++ b/packages/derived-wallet-solana/src/createSiwsEnvelope.ts
@@ -62,20 +62,18 @@ export function createSiwsEnvelopeForAptosTransaction(
  * <domain> wants you to sign in with your Solana account:
  * <base58_public_key>
  * 
- * To execute transaction <entry_function> on Aptos blockchain (<network_name>).
+ * Please confirm you explicitly initiated this request from <domain>. You are approving to execute transaction <entry_function> on Aptos blockchain (<network_name>).
  * 
  * Nonce: <aptos_txn_digest>
  */
 export function createSolanaSignMessageStatementForAptosTransaction(args:{accountAddress:string, signingMessageDigest: HexInput, rawTransaction: AnyRawTransaction  } ) {
   const {accountAddress, signingMessageDigest, rawTransaction} = args;
-  const entryFunctionName = getEntryFunctionName(rawTransaction.rawTransaction.payload);
-  const humanReadableEntryFunction = entryFunctionName ? `${entryFunctionName}` : '';
-
-  const chainId = rawTransaction.rawTransaction.chain_id.chainId;
-  const chainName = getChainName(chainId);
   
   const domain = window.location.host;
   const digestHex = Hex.fromHexInput(signingMessageDigest).toString();
 
-  return `${domain} wants you to sign in with your Solana account:\n${accountAddress}\n\nTo execute transaction ${humanReadableEntryFunction} on Aptos blockchain (${chainName}).\n\nNonce: ${digestHex}`
+  const messageStatementPrefix = `${domain} wants you to sign in with your Solana account:\n${accountAddress}`;
+  const messageStatementBody = createTransactionStatement(rawTransaction);
+
+  return `${messageStatementPrefix}\n\n${messageStatementBody}\n\nNonce: ${digestHex}`
 }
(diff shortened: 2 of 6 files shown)


packages/derived-wallet-solana/src/createSiwsEnvelope.ts
```diff
@@ -1,6 +1,8 @@
 import {
   createStructuredMessageStatement,
   createTransactionStatement,
+  getChainName,
```
hardsetting: nit: not needed, remove?

packages/derived-wallet-base/src/envelope.ts
```diff
@@ -9,7 +9,7 @@ import { StructuredMessage } from './StructuredMessage';
 /**
  * Attempt to convert the specified chainId into a human-readable identifier.
  */
-function getChainName(chainId: number) {
+export function getChainName(chainId: number) {
```
hardsetting: nit: revert as it's not needed

packages/derived-wallet-solana/src/createSiwsEnvelope.ts
```diff
@@ -51,3 +53,27 @@ export function createSiwsEnvelopeForAptosTransaction(
...
+  const {accountAddress, signingMessageDigest, rawTransaction} = args;
+  
+  const domain = window.location.host;
+  const digestHex = Hex.fromHexInput(signingMessageDigest).toString();
+
+  const messageStatementPrefix = `${domain} wants you to sign in with your Solana account:\n${accountAddress}`;
+  const messageStatementBody = createTransactionStatement(rawTransaction);
+
+  return `${messageStatementPrefix}\n\n${messageStatementBody}\n\nNonce: ${digestHex}`
+}
```
hardsetting: We can use the solana kit to build the message from the siws input

packages/derived-wallet-solana/src/signAptosTransaction.ts
```diff
@@ -39,38 +37,67 @@ export async function signAptosTransactionWithSolana(input: SignAptosTransaction
...
-    domain
-  });
 
-  const response = await wrapSolanaUserResponse(solanaWallet.signIn!(siwsInput));
-  return mapUserResponse(response, (output): AccountAuthenticator => {
-    if (output.signatureType && output.signatureType !== 'ed25519') {
-      throw new Error('Unsupported signature type');
-    }
+  // Prioritize SIWS if available
+  if(solanaWallet.signIn){
```
hardsetting, line 49: nit: there should be spaces, are you running prettier / linting?

packages/derived-wallet-solana/src/signAptosTransaction.ts
```diff
@@ -39,38 +37,67 @@ export async function signAptosTransactionWithSolana(input: SignAptosTransaction
...
+      }
+  
+      // The wallet might change some of the fields in the SIWS input, so we
+      // might need to include the finalized input in the signature.
+      // For now, we can assume the input is unchanged.
+      const signature = new Ed25519Signature(output.signature);
+  
+      return createAccountAuthenticatorForSolanaTransaction(signature, solanaPublicKey, domain, authenticationFunction, signingMessageDigest);
+    });
+  }else if(solanaWallet.signMessage){
```
hardsetting, line 64: nit: spaces

hardsetting, line 64: is the `signMessage` feature optional? :o

packages/derived-wallet-solana/src/signAptosTransaction.ts
```diff
@@ -39,38 +37,67 @@ export async function signAptosTransactionWithSolana(input: SignAptosTransaction
...
+      // Solana signMessage standard always returns a Ed25519 signature type
+      const signature = new Ed25519Signature(output);
 
-    // The wallet might change some of the fields in the SIWS input, so we
-    // might need to include the finalized input in the signature.
-    // For now, we can assume the input is unchanged.
-    const signature = new Ed25519Signature(output.signature);
+      return createAccountAuthenticatorForSolanaTransaction(signature, solanaPublicKey, domain, authenticationFunction, signingMessageDigest);
+    });
+  }else{
```
hardsetting, line 74: nit: spaces

packages/derived-wallet-solana/src/signAptosTransaction.ts
```diff
@@ -39,38 +37,67 @@ export async function signAptosTransactionWithSolana(input: SignAptosTransaction
...
+      // might need to include the finalized input in the signature.
+      // For now, we can assume the input is unchanged.
+      const signature = new Ed25519Signature(output.signature);
+  
+      return createAccountAuthenticatorForSolanaTransaction(signature, solanaPublicKey, domain, authenticationFunction, signingMessageDigest);
+    });
+  }else if(solanaWallet.signMessage){
+    // Fallback to signMessage if SIWS is not available
+    const signMessageInput = createSolanaSignMessageStatementForAptosTransaction({accountAddress: solanaPublicKey.toBase58(), signingMessageDigest, rawTransaction});
+    const signMessageInputBytes = new TextEncoder().encode(signMessageInput);
```
hardsetting: Add the following above
```typescript
import { createSignInMessage } from '@solana/wallet-standard-util';
```

Then replace with
```suggestion
    const signMessageInputBytes = createSignInMessage(siwsInput);
```

which btw internally looks like this:
```typescript
export function createSignInMessage(input: SolanaSignInInputWithRequiredFields): Uint8Array {
    const text = createSignInMessageText(input);
    return new TextEncoder().encode(text);
}
```

hardsetting: then we can remove `createSolanaSignMessageStatementForAptosTransaction` altogether I think?

hardsetting (commented):
- we can use the built-in `createSignInMessage` instead of re-defining it
- let's run the linter, formatting looks off

other than that, lgtm!

packages/derived-wallet-solana/src/signAptosTransaction.ts
0xmaayan: yes 🤦‍♀️

packages/derived-wallet-solana/src/signAptosTransaction.ts
0xmaayan: nice!

0xmaayan:
> * we can use the built-in `createSignInMessage` instead of re-defining it
> * let's run the linter, formatting looks off
> 
> other than that, lgtm!

1. Done
2. This is a big change, fmt and lint is very bad - I suggest we let this in, and then create a PR only for fmt and lint? like https://github.com/aptos-labs/aptos-wallet-adapter/pull/552

packages/derived-wallet-solana/src/signAptosTransaction.ts
```diff
@@ -46,31 +45,58 @@ export async function signAptosTransactionWithSolana(input: SignAptosTransaction
...
+  solanaPublicKey: SolanaPublicKey, 
+  domain: string, 
+  authenticationFunction: string, 
+  signingMessageDigest: Uint8Array
+) : AccountAuthenticator {
+  // Serialize the signature with the signature type as the first byte.
+  const serializer = new Serializer();
+  serializer.serializeU8(SIGNATURE_TYPE);
+  serializer.serializeBytes(signature.toUint8Array());
+  const abstractSignature = serializer.toUint8Array();
```
hardsetting, line 91: should probably define a class for this in the future :)

```typescript
export class SolanaAsbtractSignature {
  static readonly DEFAULT_TYPE = 0;
  constructor(data: Uint8Array) { ... }
  serialize(...) { }
}
```

unrelated to this pr

hardsetting (approved):

PR #561, opened 2025-04-24
X-chain wallet support for Ethereum

0xmaayan:
Add the account derivable logic for EVM wallets, following https://github.com/aptos-labs/aptos-core/pull/16367

The adapter uses evm `signMessage` (i.e `personal_sign`) to sign a message with the wallet. If the wallet supports SIWE we can expect domain verification, otherwise the default signing mechanism will be used - this is based on the wallet implementation.


apps/nextjs-x-chain/src/components/WalletProvider.tsx
```diff
@@ -10,7 +10,7 @@ import { useAutoConnect } from "./AutoConnectProvider";
 import { useToast } from "./ui/use-toast";
 
 // Temporarily disabling this, as there's no on-chain support
```
hardsetting: remove comment?

packages/derived-wallet-ethereum/src/EIP1193DerivedPublicKey.ts
```diff
@@ -57,13 +52,13 @@ export class EIP1193DerivedPublicKey extends AccountPublicKey {
     return this._authKey;
   }
 
-  verifySignature({ message, signature }: VerifySignatureArgs): boolean {
+  verifySignature({ message, signature, chainId }: VerifySignatureArgsWithChainId): boolean {
```
hardsetting: The `verifySignature` function needs to have the same signature as `AccountPublicKey`.
I'm surprised this even works, considering `chainId` is not optional :o

The chainId can't be passed externally, and needs to be either included in the signature (as I did previously) or determined from the signing message. Or maybe available as input parameter to the public key?

If the chainId is the Aptos chainId in this context, my suggestion would be to derive it from the signing message:
- if it's a transaction signing message, use the chainId included in the transaction
- if it's a message, either use the included chainId if specified, or use "0" (given that it's a required field for SIWE)

packages/derived-wallet-ethereum/src/EIP1193DerivedPublicKey.ts
```diff
@@ -89,7 +84,9 @@ export class EIP1193DerivedPublicKey extends AccountPublicKey {
   }
 
   async verifySignatureAsync(args: { aptosConfig: AptosConfig, message: HexInput, signature: Signature }): Promise<boolean> {
-    return this.verifySignature({message: args.message, signature: args.signature});
+    const aptos = new Aptos(args.aptosConfig);
+    const chainId = await aptos.getChainId();
+    return this.verifySignature({message: args.message, signature: args.signature, chainId});
```
hardsetting: see my comment above

packages/derived-wallet-ethereum/src/EIP1193DerivedPublicKey.ts
```diff
@@ -89,7 +84,9 @@ export class EIP1193DerivedPublicKey extends AccountPublicKey {
   }
 
   async verifySignatureAsync(args: { aptosConfig: AptosConfig, message: HexInput, signature: Signature }): Promise<boolean> {
```
hardsetting: What is this method?

packages/derived-wallet-ethereum/src/EIP1193DerivedPublicKey.ts
hardsetting: but again, my recommendation would be to pass the Ethereum chain id as intended by the standard.
I understand there's a chance it could be confusing, but the aptos network is specified in the statement as "Mainnet" or "Testnet" and that should be unambiguous

packages/derived-wallet-ethereum/src/EIP1193DerivedSignature.ts
```diff
@@ -4,16 +4,14 @@ export class EIP1193DerivedSignature extends Signature {
   static readonly LENGTH = 65;
 
   private readonly _siweSignature: Uint8Array;
-  readonly chainId: number;
   readonly issuedAt: Date;
 
-  constructor(siweSignature: HexInput, chainId: number, issuedAt: Date) {
+  constructor(issuedAt: Date, siweSignature: HexInput) {
```
hardsetting, line 9: why is the order changed lol
"siweSignature" should be the first imho

packages/derived-wallet-ethereum/src/EIP1193DerivedSignature.ts
```diff
@@ -22,16 +20,14 @@ export class EIP1193DerivedSignature extends Signature {
   }
 
   serialize(serializer: Serializer) {
-    serializer.serializeFixedBytes(this._siweSignature);
-    serializer.serializeU32AsUleb128(this.chainId);
-    serializer.serializeU64(this.issuedAt.getTime());
+    serializer.serializeStr(this.issuedAt.toISOString());
```
hardsetting, line 23: I think using the numeric format is better. It's more efficient in terms of memory and computation.
Is this to prevent formatting the date on chain?

packages/derived-wallet-ethereum/src/EIP1193DerivedSignature.ts
```diff
@@ -22,16 +20,14 @@ export class EIP1193DerivedSignature extends Signature {
...
+    serializer.serializeBytes(this._siweSignature);
   }
 
   static deserialize(deserializer: Deserializer) {
-    const siweSignature = deserializer.deserializeFixedBytes(EIP1193DerivedSignature.LENGTH);
-    const chainId = deserializer.deserializeUleb128AsU32();
     // Number can safely contain a unix timestamp
```
hardsetting: If you're changing the data type, this comment is not needed anymore

packages/derived-wallet-ethereum/src/createSiweEnvelope.ts
```diff
@@ -20,7 +20,7 @@ function createSiweEnvelope(input: CreateSiweEnvelopeInput & { statement: string
   return createSiweMessage({
     address: ethereumAddress,
     domain: window.location.host,
-    uri: window.location.origin,
+    uri: window.location.host,
```
hardsetting, line 23: I think `uri` requires the full URI, not just the host, did you get a chance to test this?

packages/derived-wallet-ethereum/src/signAptosTransaction.ts
```diff
@@ -1,13 +1,15 @@
...
   AccountAuthenticatorAbstraction,
   AnyRawTransaction,
   generateSigningMessageForTransaction,
   hashValues,
+  Hex,
+  Serializer,
 } from '@aptos-labs/ts-sdk';
 import { UserResponse } from "@aptos-labs/wallet-standard";
-import { BrowserProvider, Eip1193Provider } from 'ethers';
+import { BrowserProvider, Eip1193Provider, keccak256,getBytes, id, SigningKey, toUtf8Bytes, verifyMessage } from 'ethers';
```
hardsetting: let's remove unused imports please

packages/derived-wallet-ethereum/src/signAptosTransaction.ts
```diff
@@ -53,12 +55,22 @@ export async function signAptosTransactionWithEthereum(input: SignAptosTransacti
...
+
+    // Serialize the abstract public key.
+    const abstractPublicKey = new DerivableAbstractPublicKey(ethereumAddress, window.location.host);
+    
     return new AccountAuthenticatorAbstraction(
       authenticationFunction,
       signingMessageDigest,
-      authenticator,
+      abstractSignature,
+      abstractPublicKey.bcsToBytes()
```
hardsetting, line 72: I think this should be the account identity, and it shouldn't include the authentication function?
something like (ethereum_address, domain)

packages/derived-wallet-ethereum/src/EIP1193DerivedPublicKey.ts
0xmaayan: For this PR, I am mostly focus on the sign transaction message action.

For signing a message, I think we need to use the default sign message format without following the SIWE format. 

Since this whole data is mostly relevant to the sign transaction message action, I dont think we should apply it for the default sign message, i.e we should only pass in the dapp message it wants the wallet to sign, leaving out all the SIWE related fields.

image

packages/derived-wallet-ethereum/src/EIP1193DerivedPublicKey.ts
0xmaayan: was added in a different PR https://github.com/aptos-labs/aptos-wallet-adapter/pull/522

packages/derived-wallet-ethereum/src/EIP1193DerivedSignature.ts
0xmaayan: it is mostly to match the on chain function where it first deserialize the `issuedAt` (as String) and then the `siweSignature` (as vector<u8>)

packages/derived-wallet-ethereum/src/createSiweEnvelope.ts
0xmaayan: yes it works. Also it is to able to reconstruct the message on chain. sending something like `https://domain.com` is not reliable as on chain we dont now the full URI (i.e `https` or `http`)

packages/derived-wallet-ethereum/src/signAptosTransaction.ts
0xmaayan: wdym?

`abstractPublicKey ` is `DerivableAbstractPublicKey(ethereumAddress, window.location.host)` - derives by the (ethereum_address, domain).

Maybe naming is confusing. It was used to be called `account_identity` but was changed to `abstractPublicKey` on Move side (and should be updated accordingly on the AIP https://aptos-org.slack.com/archives/C08DNU2PE9W/p1743783702792199?thread_ts=1743773517.101939&cid=C08DNU2PE9W )

packages/derived-wallet-ethereum/src/EIP1193DerivedSignature.ts
0xmaayan: So, the wallet produces the signature with the date as a string, `issuedAt.toISOString()`. Then the wallet signs it. The thing is that EVM adds the message.length to the signature, so `issuedAt.toISOString()` is different in length then `issuedAt.getTime()`, and then it is almost impossible to reconstruct the message on chain given `issuedAt.getTime()` where the signature really uses `issuedAt.toISOString()`.

Unless there is a way to convert the Date milliseconds (u64) to Date iso string (String) on move.

packages/derived-wallet-ethereum/src/EIP1193DerivedPublicKey.ts
hardsetting: Yeah that's ok, we can focus on transaction signatures here if needed, but my comment above still applies.

I'm saying that the parent class expects a `verifySignature` with a specific function signature, and the method in the child class has a different function signature (it's adding a non-optional `chainId` parameter)

Maybe it's not relevant to enabling on-chain transaction authentication, but it's important we maintain code correctness

packages/derived-wallet-ethereum/src/EIP1193DerivedSignature.ts
hardsetting: Ok that's fair, if there's no easy way to convert the date format, just passing the string would be easier 👍🏼

packages/derived-wallet-ethereum/src/createSiweEnvelope.ts
hardsetting: This is what an URI looks like:
https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Example_URIs

So we're incorrectly adopting the standard.
As long as it doesn't break for wallets, we can roll with it.

But we can also just assume it's `https` to maintain correctness. i.e. `URI: https://${window.location.host}`
That should always be the case anyway, the only exception being when developing locally.
In that case, the only weridness would be seeing
`URI: https://localhost:3000` instead of `URI: http://localhost:3000`
but that's ok as URI is not verified (unlike the domain)

up to you to decide which path to take

packages/derived-wallet-ethereum/src/signAptosTransaction.ts
hardsetting: what I was trying to say is, the only information that needs to be serialized and included in the authenticator under `abstract_public_key` is what used to be called `account_identity`, in this case the concatenation of `solanaPublicKey` and `domain`.

If i understand correctly, we're passing the authentication function twice. Once directly in the authenticator as `authenticationFunction` and once encoded in the public key when calling `abstractPublicKey.bcsToBytes()`.
This should not be the case

hardsetting: mm nevermind the serializations of `DerivableAbstractPublicKey` and `SolanaDerivedPublicKey` are different.
What's confusing is `DerivableAbstractPublicKey` is not actually a public key, just a wrapper for `[identity, domain]`. I thought `SolanaDerivedPublicKey` inherited from it.

The code should be fine as-is. Maybe we can revise the naming and structure later

packages/derived-wallet-ethereum/src/EIP1193DerivedPublicKey.ts
0xmaayan: > I'm saying that the parent class expects a verifySignature with a specific function signature, and the method in the child class has a different function signature (it's adding a non-optional chainId parameter)

I already reverted it

packages/derived-wallet-ethereum/src/signAptosTransaction.ts
0xmaayan: I think `SolanaDerivedPublicKey` is a bit confusing, because it derives the daa address, tho in the context of solana **it is** the public key.

`DerivableAbstractPublicKey` matches the on-chain name, so I would like to keep it this way. I am fine with naming it `AbstractPublicKey`

hardsetting (approved):

PR #568, opened 2025-05-02
Add the URI scheme to the message URI field and EIP1193DerivedSignature

0xmaayan:
We use the `domain` as the `uri` in the message for the EVM dAA, but while it works locally (i.e `localhost:3001`) it is misleading since in production (i.e `aptos-labs.github.io`) it fails on the wallet verification because it does not follow the uri format which is `<scheme>:<path>`.

Since we need to construct the message the user signed on on-chain, and need to include the `scheme`, we add it to the abstract signature so we can extract it from there and use it in the message we build on chain.

on chain PR https://github.com/aptos-labs/aptos-core/pull/16524

image

### DIFF
--- a/packages/derived-wallet-ethereum/src/EIP1193DerivedSignature.ts
+++ b/packages/derived-wallet-ethereum/src/EIP1193DerivedSignature.ts
@@ -7,17 +7,17 @@ export class EIP1193DerivedSignature extends Signature {
   private readonly _siweSignature: Uint8Array;
   // The date and time when the signature was issued
   readonly issuedAt: Date;
-  // The origin of the message, e.g. the domain of the website that requested the signature
-  readonly origin: string;
+  // The scheme in the URI of the message, e.g. the scheme of the website that requested the signature (http, https, etc.)
+  readonly scheme: string;
 
-  constructor(origin: string, issuedAt: Date, siweSignature: HexInput) {
+  constructor(scheme: string, issuedAt: Date, siweSignature: HexInput) {
     super();
     this._siweSignature = Hex.fromHexInput(siweSignature).toUint8Array();
     if (this._siweSignature.length !== EIP1193DerivedSignature.LENGTH) {
-      throw new Error('Expected signature length to be 65 bytes');
+      throw new Error(`Expected signature length to be ${EIP1193DerivedSignature.LENGTH} bytes`);
     }
     this.issuedAt = issuedAt;
-    this.origin = origin;
+    this.scheme = scheme;
   }
 
   get siweSignature() {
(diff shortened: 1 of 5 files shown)


packages/derived-wallet-ethereum/src/EIP1193DerivedSignature.ts
```diff
@@ -3,30 +3,37 @@ import { Deserializer, Hex, HexInput, Serializer, Signature } from '@aptos-labs/
 export class EIP1193DerivedSignature extends Signature {
   static readonly LENGTH = 65;
 
+  // The signature of the message
   private readonly _siweSignature: Uint8Array;
+  // The date and time when the signature was issued
   readonly issuedAt: Date;
+  // The scheme in the URI of the message, e.g. the scheme of the website that requested the signature (http:, https:, etc.)
+  readonly scheme: string;
```
hardsetting: The scheme should not contain the semicolon. That's just something [the `URL` class does and calls "protocol"](https://url.spec.whatwg.org/#dom-url-protocol:~:text=The%20protocol%20getter%20steps%20are%20to%20return%20this%E2%80%99s%20URL%E2%80%99s%20scheme%2C%20followed%20by%20U%2B003A%20(%3A).)

packages/derived-wallet-ethereum/src/signAptosMessage.ts
```diff
@@ -72,7 +72,7 @@ export async function signAptosMessageWithEthereum(input: SignAptosMessageWithEt
   const response = await wrapEthersUserResponse(ethereumAccount.signMessage(siweMessage));
 
   return mapUserResponse(response, (siweSignature) => {
-    const signature = new EIP1193DerivedSignature(issuedAt, siweSignature);
+    const signature = new EIP1193DerivedSignature(window.location.protocol, issuedAt, siweSignature);
```
hardsetting:
```suggestion
    const scheme = window.location.protocol.slice(0, -1);
    const signature = new EIP1193DerivedSignature(scheme, issuedAt, siweSignature);
```

packages/derived-wallet-ethereum/src/signAptosTransaction.ts
```diff
@@ -57,8 +63,8 @@ export async function signAptosTransactionWithEthereum(input: SignAptosTransacti
 
     // Serialize the signature with the signature type as the first byte.
     const serializer = new Serializer();
-    serializer.serializeU8(0);
-    const signature = new EIP1193DerivedSignature(issuedAt, siweSignature);
+    serializer.serializeU8(SIGNATURE_TYPE);
+    const signature = new EIP1193DerivedSignature(window.location.protocol, issuedAt, siweSignature);
```
hardsetting:
```suggestion
    const scheme = window.location.protocol.slice(0, -1);
    const signature = new EIP1193DerivedSignature(scheme, issuedAt, siweSignature);
```

packages/derived-wallet-ethereum/src/createSiweEnvelope.ts
```diff
@@ -20,7 +20,7 @@ function createSiweEnvelope(input: CreateSiweEnvelopeInput & { statement: string
   return createSiweMessage({
     address: ethereumAddress,
     domain: window.location.host,
-    uri: window.location.host,
+    uri: window.location.protocol + '//' + window.location.host,
```
hardsetting: we can just `window.location.origin` here.

Maybe the better thing to do would be to have `domain` and `scheme` passed separately as inputs so that the message is built from its components, just like on chain.

But not needed now and outside of the scope of this PR probably

hardsetting (commented):
See my comment about the semicolon.
Sorry for being nitpicky, but I think it's important to be precise with standards :(

packages/derived-wallet-ethereum/src/EIP1193DerivedSignature.ts
hardsetting: thoughts on making this optional?
And when `undefined`, it will assumed to be `https`.
This way we have full coverage of schemes, but the signature is as short as it can be since the main use case is https and will occur like 90%+ times

hardsetting (approved):

PR #572, opened 2025-05-06
Use the default cross-chain wallet's signMessage API for when signing a regular message

0xmaayan:
- Update EVM wallets to use default `personal_sign` API when signing a regular message
- Update Solana wallets to use default `signMessage` API when signing a regular message

https://github.com/user-attachments/assets/bfcb518e-44dd-4e86-8676-9bd09cbba6cf

### DIFF
--- a/packages/derived-wallet-ethereum/src/EIP1193DerivedPublicKey.ts
+++ b/packages/derived-wallet-ethereum/src/EIP1193DerivedPublicKey.ts
@@ -18,8 +18,8 @@ import {
 import { verifyMessage as verifyEthereumMessage } from "ethers";
 import { createSiweEnvelopeForAptosTransaction } from "./createSiweEnvelope";
 import {
-  EIP1193DerivedSignature,
-  EIP1193Signature,
+  EIP1193SiweSignature,
+  EIP1193PersonalSignature,
 } from "./EIP1193DerivedSignature";
 import { EthereumAddress } from "./shared";
 
@@ -67,14 +67,14 @@ export class EIP1193DerivedPublicKey extends AccountPublicKey {
     let messageBytes: Uint8Array | string;
     // Handle structured message, i.e. a message signed withAptosSignMessageInput
     if (parsedSigningMessage.type === "structuredMessage") {
-      if (!(signature instanceof EIP1193Signature)) return false;
+      if (!(signature instanceof EIP1193PersonalSignature)) return false;
 
       messageBytes = encodeStructuredMessage(
         parsedSigningMessage.structuredMessage
       );
     } else {
       // Handle transaction
-      if (!(signature instanceof EIP1193DerivedSignature)) return false;
+      if (!(signature instanceof EIP1193SiweSignature)) return false;
       const { issuedAt } = signature;
       const signingMessageDigest = hashValues([message]);
       // Obtain SIWE envelope for the signing message
--- a/packages/derived-wallet-ethereum/src/EIP1193DerivedPublicKey.ts
+++ b/packages/derived-wallet-ethereum/src/EIP1193DerivedPublicKey.ts
@@ -1,10 +1,10 @@
 import {
   computeDerivableAuthenticationKey,
+  encodeStructuredMessage,
   parseAptosSigningMessage,
 } from "@aptos-labs/derived-wallet-base";
 import {
   AccountPublicKey,
-  Aptos,
   AptosConfig,
   AuthenticationKey,
   Deserializer,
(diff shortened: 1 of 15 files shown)


packages/derived-wallet-ethereum/src/EIP1193DerivedPublicKey.ts
```diff
@@ -59,39 +59,43 @@ export class EIP1193DerivedPublicKey extends AccountPublicKey {
...
-            chainId:
-              parsedSigningMessage.rawTransaction.rawTransaction.chain_id
-                .chainId,
-          });
-
-    const recoveredAddress = verifyEthereumMessage(siweMessage, siweSignature);
+    let messageBytes: Uint8Array | string;
+    // Handle structured message, i.e. a message signed withAptosSignMessageInput
+    if (parsedSigningMessage.type === "structuredMessage") {
+      if (!(signature instanceof EIP1193Signature)) return false;
```
hardsetting: this was previously at the top.
any reason for it to be moved?

packages/derived-wallet-ethereum/src/EIP1193DerivedSignature.ts
```diff
@@ -6,39 +6,54 @@ import {
...
+  static deserialize(deserializer: Deserializer) {
+    const signature = deserializer.deserializeBytes();
+    return new EIP1193Signature(signature);
+  }
+}
+
+/**
+ * A class representing a signature of a message signed with EIP1193 and following the SIWE standard
+ */
+export class EIP1193DerivedSignature extends EIP1193Signature {
```
hardsetting: why is a second class being defined?
any reason for not using a single class?

packages/derived-wallet-ethereum/src/signAptosMessage.ts
```diff
@@ -4,20 +4,18 @@ import {
...
-import { createSiweEnvelopeForAptosStructuredMessage } from "./createSiweEnvelope";
 import { EIP1193DerivedPublicKey } from "./EIP1193DerivedPublicKey";
-import { EIP1193DerivedSignature } from "./EIP1193DerivedSignature";
+import { EIP1193Signature } from "./EIP1193DerivedSignature";
 import { EthereumAddress, wrapEthersUserResponse } from "./shared";
 
 export interface StructuredMessageInputWithChainId
   extends StructuredMessageInput {
-  chainId: number;
+  chainId?: number;
```
hardsetting: you can just use the original StructuredMessageInput at that point

packages/derived-wallet-ethereum/src/EIP1193DerivedSignature.ts
0xmaayan: An `EIP1193Signature` is needed for when [signing a message ](https://github.com/aptos-labs/aptos-wallet-adapter/pull/572/files#diff-a45fd8bd4a291063ae3168c105c1d40f1534684bdc323da28afb774895ab9ab3R72) - this does not need the `issuedAt` and `scheme`.

An `EIP1193DerivedSignature` is needed for signing a transaction message - this requires the `issuedAt` and `scheme`.

We could make both `issuedAt` and `scheme` optional and use one class, but I prefer to have them required as they are required for signing a transaction message.

So we now have 2 classes, one defines an `EIP1193Signature` and another one `EIP1193DerivedSignature` that extends `EIP1193Signature`.

packages/derived-wallet-ethereum/src/EIP1193DerivedPublicKey.ts
0xmaayan: Because it can be either an `EIP1193Signature` or `EIP1193DerivedSignature`

See https://github.com/aptos-labs/aptos-wallet-adapter/pull/572#discussion_r2082468779

packages/derived-wallet-ethereum/src/EIP1193DerivedSignature.ts
hardsetting: Oh, I see.
This is a bit confusing and I wish we had a single class for signatures :(
but I guess it's easier this way at the moment

hardsetting: I agree with you that the fields should be required, cause they are for transactions

hardsetting: How about renaming them `EIP1193PersonalSignature` and `EIP1193SiweSignature`?
So the distinction is a bit clearer?

packages/derived-wallet-ethereum/src/EIP1193DerivedPublicKey.ts
hardsetting: Oh I thought it was the same class (names are very similar)

hardsetting (approved):
See my comment regarding possibly renaming the two classes, but otherwise approving to unblock

packages/derived-wallet-ethereum/src/EIP1193DerivedSignature.ts
hardsetting: IIUC[ SIWE is still built on Personal signatures](https://docs.login.xyz/general-information/siwe-overview/eip-4361#specification) (i.e. EIP-191), so the class hierarchy still makes sense

PR #581, opened 2025-05-13
Add `isAptosNativeWallet` prop to the wallet, update the x chain wallet…

0xmaayan:
… demo dapp to support different flows for different netwroks


packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -251,6 +252,8 @@ export class WalletCore extends EventEmitter<WalletCoreEvents> {
         }
 
         wallet.readyState = WalletReadyState.Installed;
+        // Aptos native wallets do not have an authenticationFunction property
+        wallet.isAptosNativeWallet = !("authenticationFunction" in wallet);
```
gregnazario: probably could make this a function, since it's used a couple places

gregnazario (approved):

hardsetting (approved):

PR #586, opened 2025-05-16
Remove derived wallet packages peer dependencies from the `@aptos-labs/wallet-adapter-react` package

0xmaayan:
Having the derived wallet packages as peerDependencies in the wallet-adapter-react packages requires users to install those packages even if they dont intend to use this feature.
We also can't define those as optional peerDependencies as the react package actually imports instances from those packages.

So removing the use of the derived wallet packages from the react package and move it to be implemented on the dapp level

### DIFF
--- a/.changeset/legal-bees-ring.md
+++ b/.changeset/legal-bees-ring.md
@@ -1,0 +1,5 @@
+---
+"@aptos-labs/wallet-adapter-react": minor
+---
+
+Remove derived wallet packages peer dependecies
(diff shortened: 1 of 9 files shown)


gregnazario (approved):

hardsetting (approved):

hardsetting:
Can we maybe run the linter for the whole project in a separate PR, that we can just stamp?
There's a lot of noise in every PR now :D

PR #603, opened 2025-06-30
[wallets] Reorder wallets prioritizing Petra & OKX

gregnazario:
- Remove Mizu and T wallet which are deprecated


banool (approved):

0xmaayan (commented):
Could we remove t wallet completely? from the sdkWallets file and the package dependency 

https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-core/src/sdkWallets.ts

0xmaayan:
@gregnazario Thanks. Please run changeset to include the changes

PR #604, opened 2025-06-30
Add support for base settings when building Aptos

banool:
## Stack
- Prev: N/A
- Next: #605

## Summary
This PR adds support for the dapp providing more settings for the wallet adapter to use when building an Aptos client. This can be used to override things related to txn submission for example, such as with the new txn submission plugin support.

## Test Plan
todo


banool:
Closed in favor of https://github.com/aptos-labs/aptos-wallet-adapter/pull/606.

PR #605, opened 2025-06-30
Add support for forceSubmitTransactionWithAdapter

banool:
## Stack
- Prev: #604
- Next: N/A

## Summary
This PR adds support for letting the dapp choose to force txn submission to happen via the wallet adapter, not the wallet.

## Test Plan
todo

PR #606, opened 2025-06-30
Add support for the transactionSubmitter param

banool:
## Summary
This PR adds support for setting a transaction submitter. If given, the wallet adapter will submit txns directly using this submitter rather than via the wallet.

## Test Plan
https://github.com/aptos-labs/aptos-wallet-adapter/pull/606#issuecomment-3022409934

### DIFF
--- a/.changeset/sweet-aliens-crash.md
+++ b/.changeset/sweet-aliens-crash.md
@@ -1,0 +1,13 @@
+---
+"@aptos-labs/derived-wallet-ethereum": minor
+"@aptos-labs/derived-wallet-solana": minor
+"@aptos-labs/derived-wallet-base": minor
+"@aptos-labs/wallet-adapter-core": major
+"@aptos-labs/cross-chain-core": major
+"@aptos-labs/wallet-adapter-ant-design": major
+"@aptos-labs/wallet-adapter-mui-design": major
+"@aptos-labs/wallet-adapter-react": major
+"@aptos-labs/wallet-adapter-vue": major
+---
+
+Add support for transactionSubmitter, bump minimum TS SDK version to 3.x.x
--- a/packages/cross-chain-core/package.json
+++ b/packages/cross-chain-core/package.json
@@ -75,7 +75,7 @@
     "tweetnacl": "^1.0.3"
   },
   "peerDependencies": {
-    "@aptos-labs/ts-sdk": "^1.38.0 || ^2.0.0"
+    "@aptos-labs/ts-sdk": "^3.1.1"
   },
   "files": [
     "dist",
--- a/packages/derived-wallet-base/package.json
+++ b/packages/derived-wallet-base/package.json
@@ -32,7 +32,7 @@
     "@aptos-labs/wallet-standard": "^0.5.0"
   },
   "peerDependencies": {
-    "@aptos-labs/ts-sdk": "^1.38.0 || ^2.0.0"
+    "@aptos-labs/ts-sdk": "^3.1.1"
   },
   "devDependencies": {
     "@aptos-labs/eslint-config-adapter": "workspace:*",
--- a/packages/derived-wallet-solana/package.json
+++ b/packages/derived-wallet-solana/package.json
@@ -39,7 +39,7 @@
     "@wallet-standard/app": "^1.1.0"
   },
   "peerDependencies": {
-    "@aptos-labs/ts-sdk": "^1.38.0 || ^2.0.0"
+    "@aptos-labs/ts-sdk": "^3.1.1"
   },
   "devDependencies": {
     "@aptos-labs/eslint-config-adapter": "workspace:*",
--- a/packages/wallet-adapter-core/src/utils/helpers.ts
+++ b/packages/wallet-adapter-core/src/utils/helpers.ts
@@ -4,6 +4,7 @@ import {
   Hex,
   Network,
   NetworkToNodeAPI,
+  PluginSettings,
 } from "@aptos-labs/ts-sdk";
 import {
   NetworkInfo,
@@ -62,6 +63,10 @@ export const getAptosConfig = (
     throw new Error("Undefined network");
   }
 
+  const pluginSettings: PluginSettings = {
+    TRANSACTION_SUBMITTER: dappConfig?.transactionSubmitter,
+  };
+
   if (isAptosNetwork(networkInfo)) {
     const currentNetwork = convertNetwork(networkInfo);
 
@@ -70,11 +75,13 @@ export const getAptosConfig = (
       return new AptosConfig({
         network: currentNetwork,
         clientConfig: { API_KEY: apiKey ? apiKey[currentNetwork] : undefined },
+        pluginSettings,
       });
     }
 
     return new AptosConfig({
       network: currentNetwork,
+      pluginSettings,
     });
   }
 
@@ -91,6 +98,7 @@ export const getAptosConfig = (
       return new AptosConfig({
         network: Network.CUSTOM,
         fullnode: networkInfo.url,
+        pluginSettings,
       });
     }
   }
--- a/apps/nextjs-example/package.json
+++ b/apps/nextjs-example/package.json
@@ -11,7 +11,7 @@
     "format": "prettier --write './**/*.{ts,tsx,js,jsx,json,md,html,css}'"
   },
   "dependencies": {
-    "@aptos-labs/ts-sdk": "^2.0.0",
+    "@aptos-labs/ts-sdk": "^3.1.1",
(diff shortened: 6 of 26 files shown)


0xmaayan:
With these changes, do we still need https://github.com/aptos-labs/aptos-wallet-adapter/pull/604 and https://github.com/aptos-labs/aptos-wallet-adapter/pull/605 ?

banool:
Nope I'll close those if we're happy with this.

packages/wallet-adapter-core/src/utils/helpers.ts
```diff
@@ -62,6 +63,10 @@ export const getAptosConfig = (
     throw new Error("Undefined network");
   }
 
+  const pluginSettings: PluginSettings | undefined = dappConfig?.transactionSubmitter ? {
```
0xmaayan: nit: dont think we need all of this?
just pass something like
```
{ pluginSettings: { transactionSubmitter : dappConfig?.transactionSubmitter } }
```

banool: Both work, happy to change it.

banool: Done.

0xmaayan:
Overall LGTM. 

- You should also update the ts-sdk versions in the apps (`apps/` folder) - I think this is why build is failing. (It is annoying, we should start having this dependency globally declared)
- Need to run changeset with a major upgrade version.

.changeset/sweet-aliens-crash.md
```diff
@@ -0,0 +1,13 @@
+---
+"@aptos-labs/derived-wallet-ethereum": minor
+"@aptos-labs/derived-wallet-solana": minor
+"@aptos-labs/derived-wallet-base": minor
+"@aptos-labs/wallet-adapter-core": minor
+"@aptos-labs/cross-chain-core": minor
+"@aptos-labs/wallet-adapter-ant-design": minor
+"@aptos-labs/wallet-adapter-mui-design": minor
+"@aptos-labs/wallet-adapter-react": minor
```
0xmaayan: All those packages should be major. it will break for dapps that still use previous ts-sdk versions.
"@aptos-labs/wallet-adapter-ant-design"
"@aptos-labs/wallet-adapter-mui-design"
"@aptos-labs/wallet-adapter-react"
"@aptos-labs/wallet-adapter-vue"

packages/cross-chain-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@aptos-labs/cross-chain-core",
-  "version": "4.24.13",
+  "version": "5.0.0",
```
0xmaayan: revert as this will be handled by changeset

packages/derived-wallet-base/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@aptos-labs/derived-wallet-base",
-  "version": "0.3.0",
+  "version": "0.4.0",
```
0xmaayan: revert

packages/derived-wallet-solana/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@aptos-labs/derived-wallet-solana",
-  "version": "0.3.0",
+  "version": "0.4.0",
```
0xmaayan: revert

.changeset/sweet-aliens-crash.md
```diff
@@ -0,0 +1,12 @@
+---
+"@aptos-labs/derived-wallet-ethereum": minor
+"@aptos-labs/derived-wallet-solana": minor
+"@aptos-labs/derived-wallet-base": minor
+"@aptos-labs/wallet-adapter-core": major
+"@aptos-labs/wallet-adapter-ant-design": major
+"@aptos-labs/wallet-adapter-mui-design": major
+"@aptos-labs/wallet-adapter-react": major
+"@aptos-labs/wallet-adapter-vue": major
```
0xmaayan, line 10: I think we are missing `@aptos-labs/cross-chain-core` 😄  with a minor version bump

0xmaayan (commented):
I think we got it all 😄 Mind making sure everything works by running `pnpm run dev` and see nothing breaks in the both dapps?

.changeset/sweet-aliens-crash.md
```diff
@@ -0,0 +1,13 @@
+---
+"@aptos-labs/derived-wallet-ethereum": minor
+"@aptos-labs/derived-wallet-solana": minor
+"@aptos-labs/derived-wallet-base": minor
+"@aptos-labs/wallet-adapter-core": major
+"@aptos-labs/cross-chain-core": major
```
0xmaayan, line 6:
```suggestion
"@aptos-labs/cross-chain-core": minor
```

banool:
I notice that if I have input like this:
```
    const transaction: InputTransactionData = {
      data: {
        typeArguments: [APTOS_COIN],
        // TODO: If I don't do toString here the transaction will fail, I suppose the
        // layer that does the type conversion is not being called.
        functionArguments: [account.address, 1], // 1 is in Octas
      },
    };
```
Which then gets passed to `signAndSubmitTransaction`, which in turns gets passed to this code:
```
      const transaction = await aptos.transaction.build.simple({
        sender: this._account.address,
        data: transactionInput.data,
        options: transactionInput.options,
      });

      const signTransactionResponse = await this.signTransaction({
        transactionOrPayload: transaction,
      });
      const response = await this.submitTransaction({
        transaction,
        senderAuthenticator: signTransactionResponse.authenticator,
      });
```
Which is then eventually passed to:
```
aptos.transaction.submit.simple(transaction);
```
Which just internally calls `submitTransaction`, which calls `transactionSubmitter.submitTransaction(args)`. In our implementation, we then just call the real `submitTransaction` via `aptos.transaction.submit.simple` (by setting `transactionSubmitter: null`). 

This all sounds good, but then we get this error:
```
Type mismatch for argument 0, expected 'string | AccountAddress
```

You can fix it by building the function args like this instead:
```
functionArguments: [account.address.toString(), 1]
```

So it seems like the layer that does the arg conversion isn't working for some reason, but I'm not sure why. This problem doesn't appear if you don't use the custom transaction submitter. Putting the PR in draft while I investigate.

banool:
Okay I see that in the wallet adapter we do `this._account.address.toString()` when building the txn for the wallet:

https://github.com/aptos-labs/aptos-wallet-adapter/blob/ea84adcc7ee2a35523c48b4c0516327b5507f336/packages/wallet-adapter-core/src/WalletCore.ts#L737

but not for direct submission:

https://github.com/aptos-labs/aptos-wallet-adapter/blob/ea84adcc7ee2a35523c48b4c0516327b5507f336/packages/wallet-adapter-core/src/WalletCore.ts#L782

I'm not sure if this is a bug or not.

Update: Hmm this didn't fix it.

0xmaayan:
> Okay I see that in the wallet adapter we do `this._account.address.toString()` when building the txn for the wallet:
> 
> https://github.com/aptos-labs/aptos-wallet-adapter/blob/ea84adcc7ee2a35523c48b4c0516327b5507f336/packages/wallet-adapter-core/src/WalletCore.ts#L737
> 
> but not for direct submission:
> 
> https://github.com/aptos-labs/aptos-wallet-adapter/blob/ea84adcc7ee2a35523c48b4c0516327b5507f336/packages/wallet-adapter-core/src/WalletCore.ts#L782
> 
> I'm not sure if this is a bug or not.

its a feature, not a bug 😄 

I am still not sure if it is version mismatch or just because the browser does not know how to handle those types. Because in the browser it is shown as something like
```
0: i {value: 1000000n}
```
image

And then the sdk does not know how to handle it

banool:
I tested nextjs-example with and without the custom txn submitter using Petra and everything seems to work. I also tested nuxt-example and nextjs-x-chain and they're all almost working. Notably for the nuxt example I had to switch the function arg for account addresses to use `toString`, even when not using a custom txn submitter, which feels like a regression to me, but I suppose you're saying it's expected?

There were some things that didn't work:
- Sign transaction v1 in the nuxt example fails with this error: Cannot use 'in' operator to search for 'bytecode' in undefined
- Sign as sponsor in the nuxt example fails, it seems like it tries to mint testnet APT programmatically
- Same problem with sign as sender in the multi agent flow
- Sign and submit txn in the cross chain example fails with: Error signing and submitting transaction: TypeError: Cannot read properties of undefined

I tested and confirmed that all of these errors exist on main too, so I think they're unrelated to this PR.

0xmaayan (approved):
Ship itttt

PR #609, opened 2025-06-30
Support Withdraw flow for x-chain accounts

0xmaayan:
Add support to withdraw (aka transfer) CCTP from an Aptos derived account to an origin chain (Solana) account

- Modify the `cross-chain-core` package to support Aptos -> x-chain transfer
- X-chain dapp to use the Withdraw flow

https://github.com/user-attachments/assets/e2cbfe68-d988-48c1-b90e-1ff66b5a9bf0


apps/nextjs-x-chain/src/app/components/CCTPTransfer.tsx
```diff
@@ -192,14 +190,22 @@ export function CCTPTransfer({
...
-        setTransactionInProgress(false);
         setTransactionCompleted(true);
+        setBalanceUpdatePending(true);
+        // Use delayed refetch for cross-chain transfers - origin updates immediately, Aptos balance updates after delay
+        refetchBalancesWithDelay(8000); // 8 second delay for cross-chain processing
+
+        // Clear the pending state after the delay
+        setTimeout(() => {
+          setBalanceUpdatePending(false);
+        }, 8500);
```
kent-white, line 201: This is just an example right? Cause if this is real code we probably shouldn't do this :)

apps/nextjs-x-chain/src/app/components/CCTPWithdraw.tsx
```diff
@@ -0,0 +1,381 @@
...
+  dappConfig: { aptosNetwork: dappNetwork },
+});
+const provider = crossChainCore.getProvider("Wormhole");
+
+let mainSigner: Account;
+let sponsorAccount: Account;
+
+// Set the main signer account
+const mainSignerPrivateKey =
+  process.env.NEXT_PUBLIC_SWAP_CCTP_MAIN_SIGNER_PRIVATE_KEY ||
```
kent-white, line 48: I don't have access to the secrets but is this set for the deployed version?

apps/nextjs-x-chain/src/app/page.tsx
```diff
@@ -102,12 +107,28 @@ export default function Home() {
...
+                  <CCTPTransfer
+                    wallet={wallet}
+                    originWalletDetails={originWalletDetails}
+                  />
+                  <CCTPWithdraw
+                    wallet={wallet}
+                    originWalletDetails={originWalletDetails}
+                  />
+                </>
+              )} */}
```
kent-white: just delete?

packages/cross-chain-core/src/providers/wormhole/wormhole.ts
```diff
@@ -290,6 +311,118 @@ export class WormholeProvider
...
   }
 
+  async withdraw(
+    input: WormholeWithdrawRequest
+  ): Promise<WormholeWithdrawResponse> {
+    const { sourceChain, wallet, destinationAddress, sponsorAccount } = input;
+    logger.log("sourceChain", sourceChain);
+    logger.log("wallet", wallet);
+    logger.log("destinationAddress", destinationAddress);
+    logger.log("sponsorAccount", sponsorAccount);
```
kent-white, line 321: I don't have context, but it seems like this is a lot of logging. Why are we doing so much logging here?

kent-white (approved):

apps/nextjs-x-chain/src/app/components/CCTPTransfer.tsx
0xmaayan: yes, only the demo dapp. It is because we use Indexer to query for the account usdc balance, trying to see if we can use a view function for that

apps/nextjs-x-chain/src/app/components/CCTPWithdraw.tsx
0xmaayan: yes, we store the sponsor account in a vault

packages/cross-chain-core/src/providers/wormhole/wormhole.ts
0xmaayan: yea I found it really helps for developing, we would get rid of all of these once we have a stable version, probably after we get [this](https://github.com/aptos-labs/aptos-wallet-adapter/pull/601) in

PR #631, opened 2025-09-08
[deps] Update dependencies for security audit purposes

gregnazario:
I ran `pnpm audit` and this removes most of the audit issues


0xmaayan (approved):
The `nuxt-app` build is failing. We technically should not be blocked by it, but might worth to revert the `nuxt` [dep change](https://github.com/aptos-labs/aptos-wallet-adapter/pull/631/files#diff-2f3395ff4846924203171732b609aaa3d2f134b28262cce364978d79826bc073R22)

PR #658, opened 2025-10-12
Update Aptos ts-sdk package version

0xmaayan:
- Update Aptos ts-sdk package version, to support shelbynet api key
```
<AptosWalletAdapterProvider
      dappConfig={{
        network: Network.SHELBYNET,
        aptosApiKeys: {
           shelbynet: process.env.SHELBY_API_KE,
        }
      }}
      ....
    >
      {children}
    </AptosWalletAdapterProvider>
```
- Remove usage of msafe from demo apps
- Clean up unused imports
- Mark `msafeWalletConfig` as (soon to be) deprecated

### DIFF
--- a/packages/wallet-adapter-core/src/WalletCore.ts
+++ b/packages/wallet-adapter-core/src/WalletCore.ts
@@ -2,14 +2,9 @@ import EventEmitter from "eventemitter3";
 import {
   AccountAddress,
   AccountAuthenticator,
-  AnyPublicKey,
-  AnyPublicKeyVariant,
   AnyRawTransaction,
   Aptos,
-  Ed25519PublicKey,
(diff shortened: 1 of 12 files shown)


packages/wallet-adapter-core/src/WalletCore.ts
```diff
@@ -115,7 +110,7 @@ export type AdapterNotDetectedWallet = Omit<
 };
 
 export interface DappConfig {
-  network: Network | 'shelbynet';
+  network: Network | "shelbynet";
```
hardsetting: since shelbynet is included in the sdk enum now, we should be able to just use Network here

hardsetting (approved):

PR #664, opened 2025-10-31
Rename aptos connect to petra web

maykonmichel:
- Bump @aptos-connect/wallet-adapter-plugin to 2.7.0
- Update AC dns to web.petra.app
- Rename aptos connect to petra web
- Release packages

### DIFF
--- a/.changeset/clever-animals-like.md
+++ b/.changeset/clever-animals-like.md
@@ -1,0 +1,8 @@
+---
+"@aptos-labs/wallet-adapter-ant-design": minor
+"@aptos-labs/wallet-adapter-mui-design": minor
+"@aptos-labs/wallet-adapter-react": minor
+"@aptos-labs/wallet-adapter-core": minor
+---
+
+Bump @aptos-connect/wallet-adapter-plugin to 2.7.0
--- a/.changeset/clever-animals-like.md
+++ b/.changeset/clever-animals-like.md
@@ -1,8 +1,0 @@
----
-"@aptos-labs/wallet-adapter-ant-design": minor
-"@aptos-labs/wallet-adapter-mui-design": minor
-"@aptos-labs/wallet-adapter-react": minor
-"@aptos-labs/wallet-adapter-core": minor
----
-
-Bump @aptos-connect/wallet-adapter-plugin to 2.7.0
--- a/packages/wallet-adapter-core/src/utils/walletSelector.ts
+++ b/packages/wallet-adapter-core/src/utils/walletSelector.ts
@@ -1,6 +1,10 @@
 import { WalletInfo } from "./types";
 import { AdapterNotDetectedWallet, AdapterWallet } from "../WalletCore";
-import { APTOS_CONNECT_BASE_URL, WalletReadyState } from "../constants";
+import {
+  APTOS_CONNECT_BASE_URL,
+  PETRA_WEB_BASE_URL,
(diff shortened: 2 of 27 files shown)


.changeset/clever-animals-like.md
```diff
@@ -0,0 +1,8 @@
+---
+"@aptos-labs/wallet-adapter-ant-design": minor
```
0xmaayan: those changes are not part of the version you already published?

packages/wallet-adapter-react/src/components/AboutAptosConnect.tsx
```diff
@@ -16,7 +16,7 @@ import { Web3Graphic } from "../graphics/Web3Graphic";
 import { HeadlessComponentProps, createHeadlessComponent } from "./utils";
 
 export const EXPLORE_ECOSYSTEM_URL =
-  "https://aptosfoundation.org/ecosystem/projects/all";
+  "https://aptosnetwork.com/ecosystem/directory/category/defi";
 
 const AboutAptosConnectContext = createContext<{
```
0xmaayan, line 21: should we also rename? also rename the file 
```suggestion
const AboutPetraWebContext = createContext<{
```

maykonmichel: In that case, should I also rename public functions like `isAptosConnectWallet`? Would you recommend introducing that change as a major breaking update or softly deprecating `isAptosConnectWallet` in favor of `isPetraWebWallet`?

packages/wallet-adapter-core/src/utils/walletSelector.ts
```diff
@@ -51,15 +51,15 @@ export function truncateAddress(address: string | undefined) {
   return `${address.slice(0, 6)}...${address.slice(-5)}`;
 }
 
-/** Returns `true` if the provided wallet is an Aptos Connect wallet. */
+/** Returns `true` if the provided wallet is an Petra Web wallet. */
 export function isAptosConnectWallet(wallet: WalletInfo | AdapterWallet) {
```
0xmaayan, line 63: I think this is not working anymore

packages/wallet-adapter-react/src/components/AboutAptosConnect.tsx
0xmaayan: I think we need to be consistent with the changes for users as well as for devs. There are user facing changes and dev facing changes, the later introduces some breaking changes. Perhaps we can break it into 2 releases, user facing changes can go into a minor version bump, and dev facing can go into major version bump

packages/wallet-adapter-mui-design/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@aptos-labs/wallet-adapter-mui-design",
-  "version": "5.2.0",
+  "version": "5.3.0",
```
0xmaayan, line 3: need to revert, this will get bumped through GH CI

0xmaayan (commented):
LGTM for a non-breaking version.

packages/wallet-adapter-mui-design/package.json
maykonmichel: This is from the version I released locally (broken version)

0xmaayan (approved):

PR #689, opened 2026-01-02
[core][react] Add support for Petra web fallback

GhostWalker562:
### Description
Add support for `fallback` wallets which allow for alternative connections to wallets that are not installed. This is useful in the case that wallet connections may need multiple wallet adapters to function. For example, the Petra connection can either route through `PetraWallet` (extension/mobile) or Petra web's `AptosConnectGenericWallet` (web).

Also added a `hideWallets` function that defaults to hiding the `AptosConnectGenericWallet` from the results so that it is not shown in the UI.

### Tests

https://github.com/user-attachments/assets/50baa9e6-f8dd-4c7b-9920-01b95623ac32

### DIFF
--- a/.changeset/bitter-cows-look.md
+++ b/.changeset/bitter-cows-look.md
@@ -2,5 +2,6 @@
 "@aptos-labs/wallet-adapter-core": minor
 ---
 
-Add `hideWallets` and `fallbacks` to `groupAndSortWallets`. `hideWallets` will now allow users optionally hide certain wallets from the results.
-`fallbacks` will allow users to specify fallback wallets for wallets that are not installed. This is used to support alternative connection methods for installed wallets, wallets with fallbacks will be moved to a new `availableWalletsWithFallbacks` list.
+Added `hideWallets` to `WalletCore` which will allow users optionally hide certain wallets from the default lists of wallets but will be available through the `hiddenWallets` property.
+
+Also added `fallbacks` to `groupAndSortWallets` which will allow users to specify fallback wallets for wallets that are not installed. This is used to support alternative connection methods for installed wallets, wallets with fallbacks will be moved to a new `availableWalletsWithFallbacks` list.
--- a/.changeset/bitter-cows-look.md
+++ b/.changeset/bitter-cows-look.md
@@ -1,0 +1,6 @@
+---
+"@aptos-labs/wallet-adapter-core": minor
+---
+
+Add `hideWallets` and `fallbacks` to `groupAndSortWallets`. `hideWallets` will now allow users optionally hide certain wallets from the results.
+`fallbacks` will allow users to specify fallback wallets for wallets that are not installed. This is used to support alternative connection methods for installed wallets, wallets with fallbacks will be moved to a new `availableWalletsWithFallbacks` list.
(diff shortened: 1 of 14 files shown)


packages/wallet-adapter-core/src/utils/walletSelector.ts
```diff
@@ -165,8 +259,10 @@ export function groupAndSortWallets(
     aptosConnectWallets,
     /** Wallets that use social login to create an account on the blockchain */
     petraWebWallets,
-    /** Wallets that are currently installed or loadable. */
+    /** Wallets that are currently installed or loadable or have a fallback wallet. */
     availableWallets: defaultWallets,
+    /** Wallets that are currently uninstalled that have a fallback wallet. */
+    availableWalletsWithFallbacks,
```
GhostWalker562, line 266: Made this a separate result for reverse compatibility since `availableWallets` is `AdapterWallet[]` and not `(AdapterWallet | AdapterNotDetectedWallet)[]`. In the next major version this can be merged with `availableWallets`

packages/wallet-adapter-core/src/utils/walletSelector.ts
```diff
@@ -17,22 +18,31 @@ export function partitionWallets(
   partitionFunction: (
     wallet: AdapterWallet | AdapterNotDetectedWallet,
   ) => boolean = isInstalledOrLoadable,
+  hideWallets: (
+    wallet: AdapterWallet | AdapterNotDetectedWallet,
+  ) => boolean = isPetraWebGenericWallet,
```
GhostWalker562: This could also be done at the `WalletCore` level but decided to do it here since it's the least intrusive? Thoughts?

apps/nextjs-example/src/components/WalletSelector.tsx
```diff
@@ -119,11 +121,15 @@ function ConnectWalletDialog({
...
-      [...wallets, ...notDetectedWallets],
-      walletSortingOptions,
-    );
+  const {
+    petraWebWallets,
+    availableWallets,
+    availableWalletsWithFallbacks,
+    installableWallets,
+  } = groupAndSortWallets([...wallets, ...notDetectedWallets], {
+    fallbacks: DEFAULT_WALLET_FALLBACKS,
```
GhostWalker562: Manual diff right now for reverse compatibility

.changeset/bitter-cows-look.md
```diff
@@ -0,0 +1,6 @@
+---
+"@aptos-labs/wallet-adapter-core": minor
+---
+
+Add `hideWallets` and `fallbacks` to `groupAndSortWallets`. `hideWallets` will now allow users optionally hide certain wallets from the results.
```
0xmaayan: We already have this kind of support, as `optInWallet` https://github.com/aptos-labs/aptos-wallet-adapter/blob/main/packages/wallet-adapter-core/src/WalletCore.ts#L182
This fway the core package does not handle the wallets the dapp hasn't opted in to, instead of changing the UI.

```suggestion
`hideWallets` will now allow users optionally hide certain wallets from the results.
```

GhostWalker562: Yeah I saw this. I'm looking to see if there's some way to handle this through `WalletCore` but it is kind of weird since we also need it to perform a fallback connection

lndgalante (approved):

.changeset/bitter-cows-look.md
GhostWalker562: Fixed, I updated it to provide this hidden wallet feature in `WalletCore`

PR #710, opened 2026-01-23
feat(derived-wallet-solana): Add SolanaDerivedAccount for programmatic signing

0xmaayan:
## Summary

Adds a SolanaDerivedAccount class to `@aptos-labs/derived-wallet-solana` for server-side and script-based signing using Solana keypairs.

### New Features

- New `SolanaDerivedAccount` class that extends `AbstractedAccount` for programmatic signing
  - Allows signing Aptos transactions using a Solana `Keypair` without user interaction
  - Uses `tweetnacl` for Ed25519 signing
  - Supports the SIWS (Sign-In With Solana) envelope format for on-chain verification

### Re-exports for Downstream Consumers

Added re-exports to simplify usage without requiring direct dependencies:
- `createSignInMessage` from `@solana/wallet-standard-util`
- `getChainName`, `getEntryFunctionName`, `createTransactionStatement` from `@aptos-labs/derived-wallet-base`

## Usage
```
import { Keypair } from "@solana/web3.js";
import { SolanaDerivedAccount } from "@aptos-labs/derived-wallet-solana";
import { Aptos, AptosConfig, Network } from "@aptos-labs/ts-sdk";

const account = new SolanaDerivedAccount({
  solanaKeypair: Keypair.generate(),
  domain: "my-dapp.com",
});

const aptos = new Aptos(new AptosConfig({ network: Network.TESTNET }));
const response = await aptos.signAndSubmitTransaction({
  signer: account,
  transaction: await aptos.transaction.build.simple({
    sender: account.accountAddress,
    data: { function: "0x1::aptos_account::transfer", functionArguments: [recipient, 100] },
  }),
});
```

---

> [!NOTE]
> Adds programmatic signing support with Solana keypairs and updates exports/docs for easier integration.
> 
> - New `SolanaDerivedAccount` (extends `AbstractedAccount`) enables server-side/script signing using a Solana `Keypair`, SIWS envelope, and `tweetnacl` for Ed25519
> - Updates `index.ts` to export `SolanaDerivedAccount`, wallet extension utilities, and re-export helper funcs (`createSignInMessage`, `getChainName`, etc.)
> - Renames default auth constant to `defaultSolanaAuthenticationFunction` and updates usages in `SolanaDerivedWallet` and `extendSolanaWallet`
> - README: adds programmatic signing guide, fee payer example, and marks Phantom/Solflare/Backpack/OKX as Mainnet-ready
> - package.json: adds `tweetnacl` dependency
> 
[...]

PR #711, opened 2026-01-25
feat(derived-wallet-ethereum): add node support for programmatic signing

0xmaayan:
## Summary

Add node support for programmatic signing with EIP1193DerivedAccount

- Added `EIP1193DerivedAccount` class for server-side/script-based signing
- Extracted reusable helper functions: `createMessageForEthereumTransaction` and `createAccountAuthenticatorForEthereumTransaction`
- Made `createSiweEnvelope` accept optional `domain` and `uri` parameters for Node.js compatibility
- Exported `defaultAuthenticationFunction` from shared module
- Added new `/node` entry point for Node.js-specific exports

---

> [!NOTE]
> Adds Node/script-friendly programmatic signing using Ethereum keypairs and refactors signing utilities.
> 
> - Introduces `EIP1193DerivedAccount` for server-side signing via ethers `Wallet`; computes derived Aptos address and signs transactions using SIWE
> - Extracts helpers: `createMessageForEthereumTransaction` and `createAccountAuthenticatorForEthereumTransaction` for Node environments
> - Enhances `createSiweEnvelope` to accept optional `domain`/`uri`; updates `EIP1193DerivedPublicKey.verifySignature` to include `scheme` and domain/uri in SIWE envelope
> - Expands package exports (including `defaultEthereumAuthenticationFunction`, `EthereumAddress`) and updates README with Node usage; adds mainnet support table and minor doc cleanups
> - Refactors `signAptosTransaction` to use new helpers and explicit env-derived `domain`/`uri`/`scheme`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d445e925c17acd7cc4a24f0f8b29ebdde50a131d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/derived-wallet-ethereum/src/EIP1193DerivedAccount.ts
```diff
@@ -0,0 +1,127 @@
+import {
```
GhostWalker562, line 1: nit: I don't think there's anything necessarily node specific in this account, having it both exposed to browser seems fine unless this practice suggestion. That being said, if somebody wants to work with adapter or private key I would assume they would naturally realize that they cannot use the adapter with this `EIP1193DerivedAccount` class.

Additionally it is somewhat inconsistent with Web3 SDKs since our `Account`'s from the ts-sdk are exposed on both browser and node environments, likewise Viem (I think with polyfills though)

packages/derived-wallet-ethereum/src/shared.ts
```diff
@@ -7,6 +7,9 @@ import { isError as isEthersError } from "ethers";
 
 export type EthereumAddress = `0x${string}`;
 
+export const defaultAuthenticationFunction =
+  "0x1::ethereum_derivable_account::authenticate";
```
GhostWalker562: nit: Might conflict with solana, might be worthwhile to rename to be more specific to ethereum

GhostWalker562 (approved):

packages/derived-wallet-ethereum/src/EIP1193DerivedAccount.ts
0xmaayan: yea.... the Solana package has the /node specific bundle because it uses a node only package. I tried to keep Solana and Ethereum structure somehow the same https://github.com/aptos-labs/aptos-wallet-adapter/pull/710/files#diff-9b23bdc738c5dfa1e50da62ac94c640fd80fadcec3a9db6f1928b9be1da701a3R8

But if you think it does not make sense for the Ethereum package, I am find with exposing both to the browser.

GhostWalker562: I am not sure if `tweetnacl` requires something special for browser support but its README it looks like it might support browsers too: https://www.npmjs.com/package/tweetnacl#system-requirements:~:text=Port%20of%20TweetNaCl%20/%20NaCl%20to%20JavaScript%20for%20modern%20browsers%20and%20Node.js.%20Public%20domain.

0xmaayan: interesting... it was failing for me. let me try again

0xmaayan: updated!

PR #728, opened 2026-02-03
Upgrade to support SDK v6, address security dependencies in examples, format code

gregnazario:
Also, upgrades ESlint


apps/nextjs-example/eslint.config.mjs
```diff
@@ -0,0 +1,8 @@
+import aptosConfig from "@aptos-labs/eslint-config-adapter";
+
+export default [
```
0xmaayan, line 3: what is this for?

apps/nextjs-example/next-env.d.ts
```diff
@@ -1,5 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
+/// <reference path="./.next/types/routes.d.ts" />
 
 // NOTE: This file should not be edited
-// see https://nextjs.org/docs/app/building-your-application/configuring/typescript for more information.
+// see https://nextjs.org/docs/app/api-reference/config/typescript for more information.
```
0xmaayan, line 6: this file should not be edited?

apps/nextjs-example/package.json
```diff
@@ -7,11 +7,11 @@
     "dev": "next dev --experimental-https --port 3000",
     "build": "next build",
     "start": "next start",
-    "lint": "next lint",
+    "lint": "eslint .",
     "format": "prettier --write './**/*.{ts,tsx,js,jsx,json,md,html,css}'"
   },
   "dependencies": {
-    "@aptos-labs/ts-sdk": "^5.1.1",
+    "@aptos-labs/ts-sdk": "^5.2.0",
```
0xmaayan: why not v6?

apps/nextjs-example/package.json
```diff
@@ -7,11 +7,11 @@
     "dev": "next dev --experimental-https --port 3000",
     "build": "next build",
     "start": "next start",
-    "lint": "next lint",
+    "lint": "eslint .",
```
0xmaayan, line 10: any specific reason for this change?

apps/nextjs-x-chain/eslint.config.mjs
```diff
@@ -0,0 +1,8 @@
+import aptosConfig from "@aptos-labs/eslint-config-adapter";
+
+export default [
```
0xmaayan, line 3: same

apps/nextjs-x-chain/next-env.d.ts
```diff
@@ -1,5 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
+/// <reference path="./.next/types/routes.d.ts" />
 
 // NOTE: This file should not be edited
-// see https://nextjs.org/docs/app/building-your-application/configuring/typescript for more information.
+// see https://nextjs.org/docs/app/api-reference/config/typescript for more information.
```
0xmaayan, line 6: same

apps/nextjs-x-chain/package.json
```diff
@@ -16,7 +16,7 @@
     "@aptos-labs/derived-wallet-solana": "workspace:^",
     "@aptos-labs/derived-wallet-sui": "workspace:^",
     "@aptos-labs/gas-station-client": "^2.0.3",
-    "@aptos-labs/ts-sdk": "^5.1.1",
+    "@aptos-labs/ts-sdk": "^5.2.0",
```
0xmaayan: same

apps/nextjs-x-chain/package.json
```diff
@@ -7,7 +7,7 @@
     "dev": "next dev --experimental-https --port 3001",
     "build": "next build",
     "start": "next start",
-    "lint": "next lint",
+    "lint": "eslint .",
```
0xmaayan, line 10: same

apps/nuxt-example/package.json
```diff
@@ -10,7 +10,7 @@
     "postinstall": "nuxt prepare"
   },
   "dependencies": {
-    "@aptos-labs/ts-sdk": "^5.0.0",
+    "@aptos-labs/ts-sdk": "^5.2.0",
```
0xmaayan: Aptos labs is not actively maintaining this app/package so prefer not to make any changes

packages/wallet-adapter-vue/package.json
```diff
@@ -43,13 +43,13 @@
     "build": "pnpm build:bundle"
   },
   "dependencies": {
-    "@aptos-labs/wallet-adapter-core": "5.0.0"
+    "@aptos-labs/wallet-adapter-core": "workspace:*"
```
0xmaayan, line 46: please revert. we are not actively maintaining this package and we use a fix version as this is the last tested and working version

apps/nextjs-example/next-env.d.ts
gregnazario: Yeah, funny thing is that Next.js put this "should not be edited"

Then changed the autogenerated string that gets generated into it

apps/nextjs-example/package.json
gregnazario: next version 16 does not allow `next lint`, you have to use eslint instead. a weird version change incompatibility

apps/nextjs-example/eslint.config.mjs
gregnazario: I let cursor update the eslint config to the latest, because it helped mitigate some of the random security changes (although they were only dev dependencies)

This is the way eslint configs are supposed to be today, and maybe at some point we can remove CJS

apps/nextjs-example/package.json
gregnazario: Let me go fix that, I guess cursor didn't like that.

gregnazario: Looks like there's a lot of breaking changes in other dependent packages... so it'll be a bit messier

0xmaayan: I think all apps/packages should be on the same version. We should probably start using `pnpm workspace catalog` (in a later PR)

0xmaayan (approved):

gregnazario:
Sigh... a lot of updates, let me see if I can fix this one

PR #765, opened 2026-02-19
feat(cross-chain-core): allow configuring expireTimestamp for Aptos transactions

### LINKED ISSUE #755: cross-chain-core: Allow configuring expireTimestamp for Aptos transactions
## Summary

Aptos `transaction.build.simple()` supports an `options.expireTimestamp` parameter, but the cross-chain-core Aptos signers (`AptosSigner` and `AptosLocalSigner`) don't expose it. For dapps with long-running bridge flows, the default transaction expiration may be too short, causing transactions to fail silently.

## Current behavior

In both `AptosSigner` and `AptosLocalSigner`, the transaction is built without any expiration configuration:

```ts
// AptosLocalSigner.signAndSendTransaction (line 104)
const txnToSign = await aptos.transaction.build.simple({
  data: payload,
  sender: wallet.accountAddress.toString(),
  withFeePayer: sponsorAccount ? true : false,
  // no expireTimestamp option
});
```

```ts
// AptosSigner.signAndSendTransaction (line 93)
const txnToSign = await aptos.transaction.build.simple({
  data: transactionData,
  sender: (await wallet.features["aptos:account"]?.account()).address.toString(),
  withFeePayer: sponsorAccount ? true : false,
  // no expireTimestamp option
});
```

There's no way for consumers to configure expiration timestamps for Aptos transactions in the bridge flow.

## Proposed API

Add a `getExpireTimestamp` callback to `CrossChainDappConfig`:

```ts
export interface CrossChainDappConfig {
  aptosNetwork: Network;
  // ... existing fields
  getExpireTimestamp?: () => number;  // Returns epoch seconds
}
```

## Suggested implementation

In both `AptosSigner` and `AptosLocalSigner`, read the config and pass it to `transaction.build.simple()`:

```ts
const expireTimestamp = crossChainCore?._dappConfig?.getExpireTimestamp?.();
const txnToSign = await aptos.transaction.build.simple({
  data: payload,
  sender: ...,
  withFeePayer: ...,
  ...(expireTimestamp ? { options: { expireTimestamp } } : {}),
});
```

### Prerequisite

This requires threading `CrossChainCore` through to the Aptos signers. Currently, `AptosSigner.signAndSendTransaction` only receives `dappNetwork: AptosNetwork`, not the full `CrossChainCore` instance. The router `Signer` already has access to `crossChainCore` but doesn't pass it to the Aptos-specific `signAndSendTransaction` functions.

## Use case

```ts
const crossChainCore = new CrossChainCore({
  dappConfig: {
    aptosNetwork: Network.MAINNET,
    getExpireTimestamp: () => Math.floor(Date.now() / 1000) + 120, // 2 minutes
  },
});
```

This was requested by an external team who had to patch the package to support custom transaction expiration for production bridge flows.

0xmaayan:
## Summary

Closes #755

Aptos `transaction.build.simple()` supports an `options.expireTimestamp` parameter, but the cross-chain-core Aptos signers didn't expose it. For dapps with long-running bridge flows, the default transaction expiration may be too short, causing transactions to fail silently.

### New `getExpireTimestamp` on `CrossChainDappConfig`

```ts
export interface CrossChainDappConfig {
  aptosNetwork: Network;
  // ...existing fields
  getExpireTimestamp?: () => number;  // Returns epoch seconds
}
```

This is a **function** (not a static number) because:
- Bridge flows span multiple transactions over minutes (burn → attestation → claim)
- Each transaction needs a fresh expiration window relative to the current time
- A static number frozen at config-creation time would go stale during long flows

### Where it's threaded

1. **`AptosSigner.signAndSendTransaction`** — now receives `crossChainCore` from the router `Signer`, reads `getExpireTimestamp` from the dapp config
2. **`AptosLocalSigner`** — accepts `getExpireTimestamp` as a constructor param, passed from `WormholeProvider.claimCCTPTransfer`
3. **Router `Signer.signAndSendTransaction`** — now passes `crossChainCore` to the Aptos signer path

### Usage

```ts
const crossChainCore = new CrossChainCore({
  dappConfig: {
    aptosNetwork: Network.MAINNET,
    getExpireTimestamp: () => Math.floor(Date.now() / 1000) + 120, // 2 min from now
  },
});
```

## Changeset

Minor — new optional config field, fully backward compatible.

---

> [!NOTE]
> **Medium Risk**
> Touches Aptos transaction build/sign/submit paths; while the new behavior is optional and validated, misconfiguration could cause transaction failures or unexpected expirations.
> 
> **Overview**
> Adds optional `getExpireTimestamp` to `CrossChainDappConfig` to supply per-transaction Aptos `options.expireTimestamp` at build time (fresh window for multi-step flows).
> 
> Threads this callback through the Aptos signing stack (`Signer` → `AptosSigner`, and `WormholeProvider` → `AptosLocalSigner`) and introduces `validateExpireTimestamp` to fail fast on invalid values. Adds unit tests covering forwarding/omission and invalid timestamp handling for both signers, plus a changeset for a minor release.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7c38a4c787d4b92b67e6a4e9729a95f0ce51fa79. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.changeset/add-expire-timestamp-config.md
+++ b/.changeset/add-expire-timestamp-config.md
@@ -1,0 +1,6 @@
+---
+"@aptos-labs/cross-chain-core": minor
+---
+
+Added `getExpireTimestamp` option to `CrossChainDappConfig` for configuring Aptos transaction expiration timestamps. The callback is invoked at transaction-build time so each transaction in a multi-step bridge flow gets a fresh expiration window. Threaded through both `AptosSigner` and `AptosLocalSigner`.
+
--- a/packages/cross-chain-core/src/CrossChainCore.ts
+++ b/packages/cross-chain-core/src/CrossChainCore.ts
@@ -29,6 +29,17 @@ import {
 export interface CrossChainDappConfig {
   aptosNetwork: Network;
   disableTelemetry?: boolean;
+  /**
+   * Returns an epoch-second timestamp used as `expireTimestamp` when building
+   * Aptos transactions. Called at transaction-build time so that each
+   * transaction in a multi-step bridge flow gets a fresh expiration window.
+   *
+   * @example
+   * ```ts
+   * getExpireTimestamp: () => Math.floor(Date.now() / 1000) + 120 // 2 minutes
+   * ```
+   */
+  getExpireTimestamp?: () => number;
   solanaConfig?: {
     rpc?: string;
     priorityFeeConfig?: {
--- a/packages/cross-chain-core/src/providers/wormhole/signers/AptosLocalSigner.ts
+++ b/packages/cross-chain-core/src/providers/wormhole/signers/AptosLocalSigner.ts
@@ -30,6 +30,7 @@ export class AptosLocalSigner<
   _wallet: Account;
   _sponsorAccount: Account | GasStationApiKey | undefined;
   _onTransactionSigned: OnTransactionSigned | undefined;
+  _getExpireTimestamp: (() => number) | undefined;
   _claimedTransactionHashes: string[] = [];
   _dappNetwork: AptosNetwork;
   constructor(
@@ -39,6 +40,7 @@ export class AptosLocalSigner<
     feePayerAccount: Account | GasStationApiKey | undefined,
     dappNetwork: AptosNetwork,
     onTransactionSigned?: OnTransactionSigned,
+    getExpireTimestamp?: () => number,
   ) {
     this._chain = chain;
     this._options = options;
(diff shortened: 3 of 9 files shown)


Alexzander-Stone (approved):
lgtm

Should we validate the expiration timestamp is within the constraints of Aptos (2 min I believe)? Or does the pre-existing check during TXN simulation/submission already cover this well enough?

gregnazario (approved):

0xmaayan:
> lgtm
> 
> Should we validate the expiration timestamp is within the constraints of Aptos (2 min I believe)? Or does the pre-existing check during TXN simulation/submission already cover this well enough?

I am not sure about the Node check. But I think even if we want to have this check on the client/sdk side, it should be in the ts-sdk and not in the adapter code

PR #785, opened 2026-03-12
Migrate formatting and linting to biome to speed up CI 

gregnazario:
Biome seems to do a much better job of speed, and finding a lot of issues.  It additionally removes a much larger dependency set from the codebase.  This is formatting it all prior to us making another release to update the underlying dependencies.


0xmaayan (approved):
lets land it asap, so other it will not get blocked by other incoming PRs

